PR #2, opened 2016-12-09
Update widget embedder semver range

SylvainCorlay:


SylvainCorlay:
This enables duplicating the embedder script tag without duplicating the rendered widgets.

Issue #6, opened 2017-01-09
work with 3rd party package yet (e.g: nglview)?


hainm:
thanks for this cool project.

I've tried with [nglview](https://github.com/arose/nglview) but got 
```bash
Code Execution failed:AttributeError: 'NoneType' object has no attribute 'session'
  e.__class__.__name__, str(e)))
```

Works fine with `ipywidgets.Text`
![capture](https://cloud.githubusercontent.com/assets/4451957/21782039/f0bfccee-d67f-11e6-8f85-65c26ed094af.png)

hainm:
this is the expected output if using `nglview`

![nglview](https://cloud.githubusercontent.com/assets/4451957/21782070/0f0cea74-d680-11e6-9bcd-2e755e0c3045.png)

hainm:
cc @arose

SylvainCorlay:
@hainm this requires ipywidgets 6.0 (recent beta or release candidate). It should work with custom widgets.

Closing this issue.

PR #9, opened 2017-01-18
Include default values

maartenbreddels:
This might also fix issue #6  I guess the default also need to be in the js side, but I guess nobody does this.

### DIFF
--- a/jupyter_sphinx/embed_widgets.py
+++ b/jupyter_sphinx/embed_widgets.py
@@ -209,7 +209,7 @@ def generic_visit_widget(self, node):
 
 def add_widget_state(app, pagename, templatename, context, doctree):
     if 'body' in context and Widget.widgets:
-        state_spec = json.dumps(Widget.get_manager_state(drop_defaults=True))
+        state_spec = json.dumps(Widget.get_manager_state(drop_defaults=False))
         Widget.widgets = {}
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Hum, that should not be necessary. The trick for your widget to work without using the default value is to provide them on the js side from the `default()` option. Numerous examples in ipywidgets, bqplot etc...

maartenbreddels:
But it is not consistent with the 'Embed widget' functionality in the notebook, that includes the whole state. 
What is actually the idea behind having the defaults at both the js and py side?

SylvainCorlay:
> But it is not consistent with the 'Embed widget' functionality in the notebook, that includes the whole state.

Actually it is consistent, except that if you don't specify default values on the js side, everything is considered different from the default value...

maartenbreddels:
Well, from a user point of view it is not consistent. I a piece of code, run it in the notebook, embed it in a html page and it works (full state is in the json). I take the same code, put it in the documentation and it doesn't work.
But I'll stop whining and add the defaults to the js side as well 😉

SylvainCorlay:
Adding the default values will help you for the backward compatibility issues too. You will be able to add new model attributes on the js side while remaining compatible with the same version of the backend.

PR #12, opened 2017-03-17
MAINT: correct readme formatting

stsievert:
Readme was mostly code, including content that was not code.


SylvainCorlay:
Good catch thanks.

PR #19, opened 2018-07-12
Overwrite IPython display earlier in the sphinx build

martinRenou:
Overwrite `IPython.display.display` earlier in the sphinx build in order to avoid other packages to import it before any `.. ipywidgets-setup:: ...`. See [this](https://github.com/bloomberg/bqplot/pull/673) discussion

### DIFF
--- a/jupyter_sphinx/embed_widgets.py
+++ b/jupyter_sphinx/embed_widgets.py
@@ -45,6 +45,7 @@ The directives have the following options::
 import os
 import json
 import warnings
+import sys
 
 from docutils import nodes
 from docutils.parsers.rst import Directive
@@ -59,6 +60,7 @@ from sphinx.util.nodes import set_source_info
 import ast
 import logging
 
+_ipython_display_module = sys.modules["IPython.display"]
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
I think that we can do something cleaner to prevent the use of the global `jupyter_sphinx_body` toggle.

- Essentially, at the top level:

```python
def nodisplay(*objs, **kwargs):
    pass

def jsphinx_display(*objs, **kwargs):
   for obj in objs:
       if isinstance(obj, Widget):
           view_spec = json.dumps(obj.get_view_spec())
           jupyter_sphinx_body.append('<script type="application/vnd.jupyter.widget-view+json">' + view_spec + '</script>')

_default_display = [nodisplay]

def _monkey_display(*args, **kwargs):
    return _default_display[0](*args, **kwargs)

sys.modules['IPython.display'] = _monkey_display

def set_custom_display(disp):
    _default_display[0] =  disp
```

so that we always go through `_monkey_display`

- In the code for widget setup:

    ```python
    set_custom_display(nodisplay)
    ```

- In the code for widget display:

    ```python
    set_custom_display(jsphinx_display)
    ```

martinRenou:
Ok, I see your point, I agree that it's cleaner like that

martinRenou:
Changes made and squashed :)

SylvainCorlay:
For reference: https://github.com/sphinx-gallery/sphinx-gallery/issues/396#issuecomment-403563125

SylvainCorlay:
Great thanks!

PR #22, opened 2018-08-16
add an extension for executing code in a jupyter kernel

jbweston:
Co-Authored-By: Anton Akhmerov <anton.akhmerov@gmail.com>

@SylvainCorlay as discussed on Gitter the other day.

You can try it out on the following sample rst file: https://pastebin.com/tG10Q9U0

The basic functionality is there:

+ execution is handled by a Jupyter kernel ~~(currently this is hard-coded to a Python kernel)~~
+ support for the following output formats: images, HTML, Latex and plaintext
+ the format selected for each output is done based on a configurable priority list
+ directive allows to configure hiding input or output
+ currently we use `nbconvert` to handle both execution and output extraction

Before this could be merged we still need to:

+ add support for jupyter widgets (implemented in a [separate PR](https://github.com/jbweston/jupyter-sphinx/pull/2) that may be discussed after this one, as there are orthogonal issues to discuss that are specific to widgets)
+ refactor selection of mimetype to display (currently it's a nested if/else)
+ add a proper module docstring

It would also be good to allow configuration of the kernel, but that can perhaps be in a future PR.

Please take a look and see if you agree with the overall structure, and if anything else stands out as something we could already improve.

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -1,7 +1,7 @@
 """Simple sphinx extension that executes code in jupyter and inserts output."""
 
 import os
-from itertools import takewhile, groupby, count
+from itertools import groupby, count
 from operator import itemgetter
 
 from sphinx.util import logging
@@ -48,6 +48,7 @@ def split_on(pred, it):
     """Split an iterator wherever a predicate is True."""
 
     counter = 0
+
     def count(x):
         nonlocal counter
         if pred(x):
@@ -113,19 +114,27 @@ def jupyter_download_role(name, rawtext, text, lineno, inliner):
     _, filetype = name.split(':')
     assert filetype in ('notebook', 'script')
     ext = '.ipynb' if filetype == 'notebook' else '.py'
-    # Filepaths need to be relative to the current working directory
-    # so that Sphinx will copy them.
-    output_dir = os.path.relpath(
-        output_directory(inliner.document.settings.env),
-        os.path.curdir
-    )
+    output_dir = sphinx_abs_dir(inliner.document.settings.env)
     download_file = text + ext
-    node = download_reference(download_file, download_file,
-                              reftarget=os.path.join(output_dir, download_file))
+    node = download_reference(
+        download_file, download_file,
+        reftarget=os.path.join(output_dir, download_file)
+    )
     return [node], []
 
 
-def cell_output_to_nodes(cell, data_priority):
+def cell_output_to_nodes(cell, data_priority, dir):
+    """Convert a jupyter cell with outputs and filenames to doctree nodes.
+
+    Parameters
+    ==========
+    cell : jupyter cell
+    data_priority : list of mime types
+        Which media types to prioritize.
+    dir : string
+        Sphinx "absolute path" to the output folder, so it is a relative path
+        to the source folder prefixed with ``/``.
+    """
     to_add = []
     for index, output in enumerate(cell.get('outputs', [])):
         output_type = output['output_type']
@@ -164,8 +173,10 @@ def cell_output_to_nodes(cell, data_priority):
                 # Sphinx treats absolute paths as being rooted at the source
                 # directory, so make a relative path, which Sphinx treats
                 # as being relative to the current working directory.
-                uri = os.path.relpath(output.metadata['filenames'][mime_type],
-                                      os.path.curdir)
+                filename = os.path.basename(
+                    output.metadata['filenames'][mime_type]
+                )
+                uri = os.path.join(dir, filename)
                 to_add.append(nodes.image(uri=uri))
             elif mime_type == 'text/html':
                 to_add.append(nodes.raw(
@@ -228,6 +239,7 @@ def write_notebook_output(notebook, output_dir, notebook_name):
         unique_key=os.path.join(output_dir, notebook_name),
         outputs={}
     )
+
     # Modifies 'resources' in-place
     ExtractOutputPreprocessor().preprocess(notebook, resources)
     # Write the cell outputs to files where we can (images and PDFs),
@@ -247,9 +259,26 @@ def output_directory(env):
     # polluting the current working directory. We don't use a
     # temporary directory, as sphinx may cache the doctree with
     # references to the images that we write
+
+    # Note: we are using an implicit fact that sphinx output directories are
+    # direct subfolders of the build directory.
     return os.path.abspath(os.path.join(
         env.app.outdir, os.path.pardir, 'jupyter_execute'
     ))
+
+
+def sphinx_abs_dir(env):
+    # We write the output files into
+    # output_directory / jupyter_execute / path relative to source directory
+    # Sphinx expects download links relative to source file or relative to
+    # source dir and prepended with '/'. We use the latter option.
+    return '/' + os.path.relpath(
+        os.path.abspath(os.path.join(
+            output_directory(env),
+            os.path.dirname(env.docname),
+        )),
+        os.path.abspath(env.app.srcdir)
+    )
 
 
 class ExecuteJupyterCells(SphinxTransform):
@@ -276,7 +305,15 @@ class ExecuteJupyterCells(SphinxTransform):
 
         for nodes in nodes_by_notebook:
             kernel_name = nodes[0]['kernel_name'] or default_kernel
-            notebook_name = nodes[0]['notebook_name'] or next(default_names)
+            notebook_name = nodes[0]['notebook_name']
+
+            if notebook_name:
+                notebook_name = os.path.join(
+                    os.path.dirname(docname), notebook_name
+                )
+            else:
+                notebook_name = next(default_names)
+
             notebook = execute_cells(
                 kernel_name,
                 [nbformat.v4.new_code_cell(node.astext()) for node in nodes],
@@ -289,7 +326,9 @@ class ExecuteJupyterCells(SphinxTransform):
             # we just wrote to; sphinx copies these when writing outputs.
             for node, cell in zip(nodes, notebook.cells):
                 output_nodes = cell_output_to_nodes(
-                    cell, self.config.jupyter_execute_data_priority
+                    cell,
+                    self.config.jupyter_execute_data_priority,
+                    sphinx_abs_dir(self.env)
                 )
                 attach_outputs(output_nodes, node)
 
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
@jbweston @akhmerov this is great! Sorry for not responding earlier, I was having meetings back to back the whole afternoon. I will be looking at this tomorrow!

SylvainCorlay:
I am looking at it now. Making a few nitpicking commands, but in general, I love the direction that this is  taking.

akhmerov:
Some points worthy to consider at this stage:

- How much flexibility do we want to allow? Is it OK to let users specify "cells" in segments? Out of order? Separately specify where the output should be plugged in?
- At which stage of sphinx pipeline should we do the processing?
- Is the current syntax choice best? Or would it be better to add a different directive that instead goes "please treat all the following code blocks as notebook cells"?

akhmerov:
Another question: let's imagine the users want to allow handling of a new mime type (say geojson). Would we be able to allow them to do it at the level of configuration?

jupyter_sphinx/execute.py
```diff
@@ -0,0 +1,373 @@
...
+        return [Cell('',
+            nodes.literal_block(
+                text='\n'.join(self.content),
+                language='ipython'
+            ),
+            hide_code=('hide-code' in self.options),
+            hide_output=('hide-output' in self.options),
+            code_below=('code-below' in self.options),
+            kernel_name=self.options.get('kernel', '').strip(),
+            new_notebook=('new-notebook' in self.options),
```
SylvainCorlay: I am not sure about the `new_notebook` option, which seems to be used as a means to make code snippets run in separate execution contexts.

I would prefer something like `restart_kernel`. Using`notebook` in the name of the options seems to expose the fact that it is creating a notebook under the hood, which is an implementation detail IMO.

We could remove the feature altogether for a first version, since this may be confusing to users. Or we could havesomething like `kernel_id` which will specify the kernel id to be created or something like this.

SylvainCorlay:
> How much flexibility do we want to allow? Is it OK to let users specify "cells" in segments? Out of > order? Separately specify where the output should be plugged in?

Maybe we should keep the package as simple and lean as possible at first before adding feature that we don't know are really needed.

> At which stage of sphinx pipeline should we do the processing?
> Is the current syntax choice best? Or would it be better to add a different directive that instead goes "please treat all the following code blocks as notebook cells"?

I like one-size-fits-all directive that you implemented better than the two directives that we have now.

> Another question: let's imagine the users want to allow handling of a new mime type (say geojson). Would we be able to allow them to do it at the level of configuration?

The case of geo+json should be handled with the `+json` suffix that catches all into a script tag. Maybe registering custom mime type renderers will be the way to go in the future.

jupyter_sphinx/execute.py
```diff
@@ -0,0 +1,373 @@
...
+    )
+    app.add_config_value(
+        'jupyter_execute_data_priority',
+        [
+            'text/html',
+            'image/svg+xml',
+            'image/png',
+            'image/jpeg',
+            'text/latex',
+            'text/plain'
```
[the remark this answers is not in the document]
akhmerov: What takes care of the rendering the script? Extra page header? What if we want to render something also in latex output?

Is it a standard behavior for various types of data? I am particularly interested in plotly, but they seem to not have a renderer that would function this way.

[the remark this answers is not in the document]
SylvainCorlay: There is a library called the HTML embedder which can be included as a script tag. This is what we use to render widgets in e.g. http://nbviewer.jupyter.org/github/jupyter-widgets/ipywidgets/blob/master/docs/source/examples/Widget%20List.ipynb

Although, it will also require the widgets state to be stored in a separate `application/vnd.jupyter.widget-view+json` script tag. I will probably open a PR adding this later.

Note that plotly is now based on Jupyter widgets (since 3.0) so if they did things right, this will also enable the rendering of plotly charts.

jupyter_sphinx/execute.py
akhmerov: I agree about the naming, indeed. I also agree that under-the-hood creation of notebooks is not a good thing to advertise.

At the same time we do want to allow the users to easily extract and run the code, and also specify the filename. Does that make sense?

So the user may need to:
* Specify the filename
* Indicate that the new kernel is needed

Not entirely sure what's the best way to translate these two parameters into options, and how to call those.

SylvainCorlay: I am not sure what you mean about the file name?

jupyter_sphinx/execute.py
```diff
@@ -0,0 +1,373 @@
...
+    )
+
+    # Modifies 'resources' in-place
+    ExtractOutputPreprocessor().preprocess(notebook, resources)
+    # Write the cell outputs to files where we can (images and PDFs),
+    # as well as the notebook file.
+    FilesWriter(build_directory=output_dir).write(
+        nbformat.writes(notebook), resources,
+        os.path.join(output_dir, notebook_name + '.ipynb')
+    )
```
SylvainCorlay: Do we necessarily need to write a notebook file to disk?

jupyter_sphinx/execute.py
akhmerov: in the `jupyter-download:...` directives we allow to download the code as a notebook or a script. So we'd need to specify a filename for it to be human-readable.

jupyter_sphinx/execute.py
akhmerov: This is in case we want to let the user download it. I think this is quite useful for more involved tutorials.

SylvainCorlay: Gotcha, although it exposes the "notebook" nature of the directive.

jupyter_sphinx/execute.py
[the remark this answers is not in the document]
akhmerov: They have `FigureWidget`, but it doesn't seem to save the actual figure data in the widget state right now. (This should go into `widget_state`, right?) They also support bundling json with the output.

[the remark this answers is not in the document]
akhmerov: Do you have a link to the HTML embedder library? I failed to find it after a quick google search.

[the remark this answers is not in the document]
SylvainCorlay: The name of the package is `@jupyter-widgets/html-manager` 
 and the code is in the ipywidgets repository. You can read more about it here:

https://ipywidgets.readthedocs.io/en/stable/embedding.html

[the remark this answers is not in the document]
akhmerov: Would this also work with scripts that have `src` tag set? Some widgets will contain a lot of data, and extracting them into a separate file would improve the page load times.

Thinking of producing output other than html: I remember there was a way to render a widget into a static image, is that right? Does it work on custom widgets? Can this be done without a browser (e.g. via an electron app)?

akhmerov:
Summary of the discussion:

- Rename the directive to `jupyter-execute`
- Extend it with `source` option allowing to include code from files (relevant for e.g. initialization code)
- Factor out the "global" options into a separate directive, named e.g. `new-kernel`, with optional options `kernel-name`, `filename` (and not `notebook-name`!), maybe even default state for showing/hiding code and source.
- Rely on nbconvert 5.4, implement querying the kernel for widget state after we have submitted all the code to it (@SylvainCorlay).
- Read the file extension off from the kernel metadata (requires nbconvert 5.4).

jbweston:
> 

    + Rename the directive to jupyter-execute
    + Extend it with source option allowing to include code from files (relevant for e.g. initialization code)
    + Factor out the "global" options into a separate directive, named e.g. new-kernel, with optional options kernel-name, filename (and not notebook-name!), maybe even default state for showing/hiding code and source.

@akhmerov I addressed these 3 points. There is https://github.com/jbweston/jupyter-sphinx/pull/2 that is aiming to address the point about widgets

akhmerov:
Awesome, let's merge that one (there are already conflicts). Then, as far as I can judge, the remaining bits will amount to cleanup and review.

jupyter_sphinx/execute.py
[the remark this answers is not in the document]
jbweston: There is now https://github.com/jbweston/jupyter-sphinx/pull/2 where I've started implementing ipywidgets support. Perhaps we should first merge this PR and then we can address that one (there are still several unanswered questions.

jupyter_sphinx/execute.py
jbweston: 603ef72 introduces a `jupyter-kernel` directive that is used to specify the kernel to use:

```
.. jupyter-kernel:: python3
    :id: my-kernel-id
```

the `id` is used as the filename and the argument to the directive is the kernel name.

jupyter_sphinx/execute.py
jbweston: I addressed the naming of the directive options elsewhere, but I believe that we should keep writing the notebooks. Are we in agreement on that?

jbweston:
@akhmerov I would prefer to merge this first and then I can point https://github.com/jbweston/jupyter-sphinx/pull/2 to this repository instead. 

IMO there's already a lot in this PR as it is, and there are actually a few more points that we need clarity on wrt. widget support (e.g. how to know what versions of the embedding JS to use).

I'd wager we'd still land https://github.com/jbweston/jupyter-sphinx/pull/2 before Sylvain wants to make a release.

jbweston:
> The case of geo+json should be handled with the +json suffix that catches all into a script tag. Maybe registering custom mime type renderers will be the way to go in the future.

Indeed, I have a branch where I'm trying this out, but I believe this change can be made in a later PR, as the output rendering is quite isolated and won't require changes in several places.

jbweston:
@SylvainCorlay I believe that we've responded to all of your comments, it'd be good to get another look over to see if there's anything we've missed.

jupyter_sphinx/execute.py
```diff
@@ -351,6 +350,11 @@ def apply(self):
                 [nbformat.v4.new_code_cell(node.astext()) for node in nodes],
                 self.config.jupyter_execute_kwargs,
             )
+
+            for node in nodes:
+                source = node.children[0]
+                source.attributes['language'] = notebook.metadata.language_info.name
```
akhmerov: The kernels (ipython most importantly) may also provide a `pygments_lexer` key, which seems better to use here.

e.g. ipython's language is `python`, but `ipython` is a better lexer.

jbweston: good point, will amend now

jbweston:
@SylvainCorlay did you get a chance to look at this?

PR #24, opened 2018-11-25
add support for rendering ipywidgets

jbweston:
Co-authored-by: Bas Nijholt <basnijholt@gmail.com>

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -17,7 +17,7 @@ from IPython.lib.lexers import IPythonTracebackLexer, IPython3Lexer
 from docutils.parsers.rst import Directive, directives
 
 import nbconvert
-from nbconvert.preprocessors.execute import ExecutePreprocessor
+from nbconvert.preprocessors.execute import executenb
 from nbconvert.preprocessors import ExtractOutputPreprocessor
 from nbconvert.writers import FilesWriter
 
@@ -483,57 +483,11 @@ def contains_widgets(notebook):
     return widgets and widgets['state']
 
 
-# TODO: Remove this once  https://github.com/jupyter/nbconvert/pull/900
-#       is merged and nbconvert 5.5 is released.
-def extract_widget_state(executor):
-    """Extract ipywidget state from a running ExecutePreprocessor"""
-    # Can only run this function inside 'setup_preprocessor'
-    assert hasattr(executor, 'kc')
-    # Only Python has kernel-side support for jupyter widgets currently
-    if language_info(executor)['name'] != 'python':
-        return None
-
-    get_widget = '''\
-        state = None
-        try:
-            import ipywidgets
-            state = ipywidgets.Widget.get_manager_state()
-        except Exception:  # Widgets are not installed in the kernel env
-            pass
-        state
-    '''
-    cell = nbformat.v4.new_code_cell(get_widget)
-    _, (output,) = executor.run_cell(cell)
-    return literal_eval(output['data']['text/plain'])
-
-
 def language_info(executor):
     # Can only run this function inside 'setup_preprocessor'
     assert hasattr(executor, 'kc')
     info_msg = executor._wait_for_reply(executor.kc.kernel_info())
     return info_msg['content']['language_info']
-
-
-# Vendored from 'nbconvert.preprocessors.executenb' with modifications
-# to extract widget state from the kernel after execution and store it
-# in the notebook metadata.
-# TODO: Remove this once  https://github.com/jupyter/nbconvert/pull/900
-#       is merged and nbconvert 5.5 is released.
-def executenb(nb, cwd=None, km=None, **kwargs):
-    """Execute a notebook and embed widget state."""
-    resources = {}
-    if cwd is not None:
-        resources['metadata'] = {'path': cwd}
-    ep = ExecutePreprocessor(**kwargs)
-    with ep.setup_preprocessor(nb, resources, km=km):
-        ep.log.info("Executing notebook with kernel: %s" % ep.kernel_name)
-        nb, resources = super(ExecutePreprocessor, ep).preprocess(nb, resources)
-        nb.metadata.language_info = language_info(ep)
-        widgets = extract_widget_state(ep)
-        if widgets:
-            nb.metadata.widgets = {WIDGET_STATE_MIMETYPE: widgets}
-
-
 
 
 def write_notebook_output(notebook, output_dir, notebook_name):
@@ -593,7 +547,7 @@ def setup(app):
     # Configuration
     app.add_config_value(
         'jupyter_execute_kwargs',
-        dict(timeout=-1, allow_errors=True),
+        dict(timeout=-1, allow_errors=True, store_widget_state=True),
         'env'
     )
     app.add_config_value(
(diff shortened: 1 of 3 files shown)


akhmerov:
Something to consider: widget state may be extremely large, and potentially block loading of the document. This is already done with the output types that support extraction.

Case in point: https://adaptive.readthedocs.io/en/latest/docs.html is 6.1 MB and loads for some 10-20 seconds using a decent speed connection.

If possible, I believe that loading the widget state as a separate request would be a better solution.

@SylvainCorlay do you think if this can be done?

jbweston:
> If possible, I believe that loading the widget state as a separate request would be a better solution.

IIRC there's already a comment about this: the html-embed library for ipywidgets reads the innerHTML of all script tags that have the appropriate type (something something 'widget-state'). This means that we have to load it in the page.

We can help the situation by making sure that we place the state JS as far down the page as possible, although of course this will still block the "loaded" event for the page until the browser has parsed the script tag...

jbweston:
Seems that loading external  widget state is on the radar for ipywidgets, but no fixed timeline: https://github.com/jupyter-widgets/ipywidgets/issues/1299

SylvainCorlay:
Absolutely.

Also the size of the state sent to the front-end can be mitigated with binary serialization in messages (rather than JSON encoding large arrays). Also, being careful to specify default values in the js side prevents its duplicated storage, which matters for small widgets repeated many times in a figure.

basnijholt:
@akhmerov actually
> Case in point: https://adaptive.readthedocs.io/en/latest/docs.html is 6.1 MB and loads for some 10-20 seconds using a decent speed connection.

The ~6 MB that you are quoting are bokeh and plotly plots, and have nothing to with the `ipywidgets`.

In the `adaptive` documentation we only use the widgets for a simple HTML table (for example [here](https://adaptive.readthedocs.io/en/latest/tutorial/tutorial.Learner1D.html#vector-output-f-n).)

akhmerov:
Indeed, I didn't realize that. Do you know why do we inline those?

basnijholt:
That's just how plotly and holoviews/bokeh do it by default, we didn't touch any of that.

jbweston:
Just to be clear: is this PR blocked on implementing loading of widget state from an external file? Or was the plan to merge this as-is and add that feature later?

akhmerov:
No, I don't think that loading widget state externally is a blocking feature for this PR.

jupyter_sphinx/execute.py
```diff
@@ -437,6 +565,31 @@ def sphinx_abs_dir(env):
...
 
 
+def builder_inited(app):
+    require_url = app.config.jupyter_sphinx_require_url
+    # 3 cases
+    # case 1: ipywidgets 6, only embed url
+    # case 2: ipywidgets 7, with require
+    # case 3: ipywidgets 7, no require
+    # (ipywidgets6 with require is not supported, require_url is ignored)
+    if has_embed:
```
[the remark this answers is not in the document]
maartenbreddels: I think there is no right approach here, it could be that the html-manager you want does not correspond to any of the versions quoted in any package installed (imagine a patch release).
I'm in favor for bumping up the required ipywidgets version to 7, drop the has_embed, and consider what ipywidgets gives us a good default, and otherwise use `app.config.jupyter_sphinx_embed_url`
@SylvainCorlay do you agree?

jupyter_sphinx/execute.py
```diff
@@ -462,6 +616,11 @@ def setup(app):
         'env',
     )
 
+    # ipywidgets config
+    require_url_default = 'https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js'
```
maartenbreddels: I don't like having duplicate default in 1 package (this is already in embed_widgets right?)

jbweston:
I think that with @basnijholt's changes this might be good to go?

basnijholt:
@maartenbreddels I've addressed your comments and made `jupyter-sphinx` depend on ipywidgets>=7.

jbweston:
@maartenbreddels is this MR ready to go now?

jupyter_sphinx/execute.py
```diff
@@ -384,6 +439,73 @@ def execute_cells(kernel_name, cells, execute_kwargs):
...
+        # not contain 'WIDGET_STATE_MIMETYPE'
+        return None
+
+
+def contains_widgets(notebook):
+    widgets = get_widgets(notebook)
+    return widgets and widgets['state']
+
+
+# TODO: Remove this once  https://github.com/jupyter/nbconvert/pull/900
```
akhmerov: This has now happened; we should rely on the default behavior of the `ExecutePreprocessor` now.

setup.py
```diff
@@ -18,7 +18,7 @@
     packages = ['jupyter_sphinx'],
     install_requires = [
         'Sphinx>=0.6',
-        'ipywidgets>=6.0.0',
+        'ipywidgets>=7.0.0',
         'IPython',
         'nbconvert>=5.4',
```
akhmerov: `nbconvert>=5.5`

akhmerov: Actually not sure about this one: embedding widgets will only work on `5.5`, however we might want to be lenient and allow for older systems that don't need to use widgets.

akhmerov:
If possible, we should try and keep `ipywidgets` an optional dependency. After a quick glance of the code, it seems indeed doable.

jupyter_sphinx/execute.py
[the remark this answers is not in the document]
akhmerov: I wonder if the configuration option should be called `app.config.jupyter_sphinx_widget_embed_url` (it's only for embedding widgets and not anything else).

akhmerov:
@maartenbreddels @SylvainCorlay I think this is ready to go, except for two unresolved discussions where your opinions would be also valuable.

setup.py
basnijholt: Shouldn't you do this now that you pushed b507db458dff2686eebd087e7e7a804fab70cad2?

akhmerov: Another option would be to introduce an `extras_require`, and make `nbconvert>=5.5` as well as `ipywidgets>=7.0.0` a prerequisite for a `widgets` extra.

The motivation for this would be a smaller dependency footprint

akhmerov: In other words: everything will still work on `nbconvert>=5.4`, except widgets won't be embedded.

SylvainCorlay:
@akhmerov @jbweston I have merged the removal of embed_widgets, so this probably needs a rebase.

setup.py
SylvainCorlay: I believe it is fine requiring `nbconvert 5.5`. The thing about allowing two versions with notably different behaviors is that it may break expectations. So for the principle of least surprise, I prefer >=5.5.

SylvainCorlay:
This looks good to me. Much cleaner than before!

I am happy to merge as soon as the PR is rebased.

setup.py
akhmerov: Done

akhmerov:
@jbweston I've added a test. Please give a final look, and I think then we're good.

@SylvainCorlay I assume you've signed off on this PR, so we'll merge once @jbweston takes a look.

SylvainCorlay:
That sounds good to me. I will enable the CI shortly.

jbweston:
LGTM

PR #27, opened 2019-03-26
WIP: add jupyter_sphinx_rel_repo_path, closes #26

### LINKED ISSUE #26: document handling packages that are not installed
running `make html` from the `repo/docs` doesn't result in `jupyter_sphinx.execute` using the code from `repo`.

This is a problem in Adaptive, whereas a bad workaround we [install master in the CI.](https://github.com/python-adaptive/adaptive/blob/master/docs/environment.yml#L20)

basnijholt:
This adds a new configurable variable `jupyter_sphinx_rel_repo_path` and solves the problem I brought up in #26.

For example, our repo structure is `repo/docs/source/conf.py` and the main `__init__.py` is in `repo/adaptive/`.

Now having `jupyter_sphinx_rel_repo_path = '../'` in `conf.py` will give the following output to `import sys; print(sys.path)`
```
['', '/Users/basnijholt/Work/adaptive', '/Users/basnijholt/miniconda3/lib/python37.zip', '/Users/basnijholt/miniconda3/lib/python3.7', '/Users/basnijholt/miniconda3/lib/python3.7/lib-dynload', '/Users/basnijholt/miniconda3/lib/python3.7/site-packages', '/Users/basnijholt/miniconda3/lib/python3.7/site-packages/IPython/extensions', '/Users/basnijholt/.ipython']
```

and before this was
```
['', '/Users/basnijholt/Work/adaptive/docs', '/Users/basnijholt/miniconda3/lib/python37.zip', '/Users/basnijholt/miniconda3/lib/python3.7', '/Users/basnijholt/miniconda3/lib/python3.7/lib-dynload', '/Users/basnijholt/miniconda3/lib/python3.7/site-packages', '/Users/basnijholt/miniconda3/lib/python3.7/site-packages/IPython/extensions', '/Users/basnijholt/.ipython']
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ File `conf.py` has two extra (optional) configuration options:
 
  * `jupyter_sphinx_require_url`: url for `require.js` (if your theme already provides this, set it to False or '')
  * `jupyter_sphinx_embed_url`: url for the embedding, if set to None (default) a proper default will be taken from the `ipywidgets.embed` module.
+ * `jupyter_sphinx_rel_repo_path`: relative path from `docs` to the repository that should be appended to `sys.path` when executing the docs.
 
 ### Misc.
 
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -209,7 +209,8 @@ class ExecuteJupyterCells(SphinxTransform):
             notebook = execute_cells(
                 kernel_name,
                 [nbformat.v4.new_code_cell(node.astext()) for node in nodes],
-                self.config.jupyter_execute_kwargs,
+                dict(self.config.jupyter_execute_kwargs,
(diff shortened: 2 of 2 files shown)


akhmerov:
A quick feedback: this feels wrong; the computational environment should be controlled by sphinx itself or by the jupyter kernels. We shouldn't be introducing a new way to do the same thing.

I'll think of an alternative solution.

akhmerov:
I confirm, I rather believe the correct way around this is to document setting the correct `PYTHONPATH`.

There *may* be a need to also set the working directory, e.g. if the users need to process some external data. This is however uncommon in building documentation and I suggest to wait with introducing extra options before we have a clear use case.

jbweston:
> I confirm, I rather believe the correct way around this is to document setting the correct PYTHONPATH

I concur; closing.

PR #28, opened 2019-04-15
add 'raises' option to jupyter-execute directive to explicitly allow cells to raise exceptions

### LINKED ISSUE #25: Exception control
Under most circumstances the code in the documentation should execute without exceptions, however sometimes showing the traceback may be useful in tutorials.

I think we should allow the user granular control declaring whether a cell may raise (leaving no exception the default), and then raising an appropriate excetion if the cell behaves opposite to what is declared.

Checking this should be as easy as checking the cell outputs.

jbweston:
Closes #25 

example:

```
.. jupyter-execute
    :raises: ZeroDevisionError

    1 / 0
```

The output will show the stacktrace as the output of the cell. If instead we had not provided `raises`, then an ExtensionError would be raised when executing the cell.

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -224,6 +224,14 @@ class ExecuteJupyterCells(SphinxTransform):
                 self.config.jupyter_execute_kwargs,
             )
 
+            # Raise error if cells raised exceptions and were not marked as doing so
+            for node, cell in zip(nodes, notebook.cells):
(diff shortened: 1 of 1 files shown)


akhmerov:
All looks good; only one suggestion though: I believe a blank `:raises:` should mean "catch all" rather than be equivalent to no `:raises:` at all. I imagine a common usage pattern would correspond to not bothering to specify the error type.

PR #29, opened 2019-04-18
strip unecessary newlines when content is loaded from a file

jbweston:
This appears when using the "read from a file" form of the `jupyter-execute` directive:

```
.. jupyter-execute:: some_file.py
    :hide-output:
```
`self.content` should seemingly be a list of strings, where each string is 1 line. Because `readlines()` returns strings with newlines at the end, we must first strip them.

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -133,7 +133,7 @@ class JupyterCell(Directive):
                 )
             try:
                 with open(filename) as f:
-                    content = f.readlines()
+                    content = [line.rstrip() for line in f.readlines()]
             except (IOError, OSError):
                 raise IOError(
                     'File {} not found or reading it failed'.format(filename)


akhmerov:
What are the unnecessary newlines?

akhmerov:
Hm, this has a side-effect of also removing trailing whitespace. An alternative `.rstrip('\n')` would on the other hand not work on files with windows-style newlines. Both are unlikely scenarios, but I consider the former less likely to cause complaints.

Therefore I'm leaning towards merging this as is. @jbweston do you agree?

jbweston:
Yes I don't know of any programming language (except for the very esoteric) where trailing whitespace is important. I'm not sure it's worth writing a function to strip line endings based on the operating system...

akhmerov:
Markdown has double trailing whitespace as a linebreak, but to the best of my knowledge there is neither a markdown kernel for jupyter nor a plan to implement one.

PR #38, opened 2019-05-01
enable parallel execution

### LINKED ISSUE #30: support parallelization
Sphinx has a [limited support for parallelization](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-j).

Because code execution via jupyter is likely a bottleneck, it's worth double-checking that this part is indeed parallelizable.

jbweston:
Closes #30.

This seems to be enough, however it still might not give the desired results. Sphinx splits the documents to be read lexicographically, which means that if the long-running documents have filenames that are close they may be assigned to the same parallel worker. This is, for example, the case with the Kwant documentation, where all the jupyter sphinx happens in files that start with `tutorial/`

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -526,4 +526,7 @@ def setup(app):
     app.add_lexer('ipythontb', IPythonTracebackLexer())
     app.add_lexer('ipython', IPython3Lexer())
 
-    return {'version': __version__}
+    return {
+        'version': __version__,
+        'parallel_read_safe': True,
+    }


akhmerov:
I also saw `parallel_write_safe`; is that important for us?

jbweston:
Freshly opened: https://github.com/sphinx-doc/sphinx/issues/6330

jbweston:
> I also saw parallel_write_safe; is that important for us?

No; our doctree transformation happens in the reading step.

akhmerov:
With the upstream issue there's nothing more to do on our side.

basnijholt:
How do I enable the parallelization of the execution of the scripts?

Here https://github.com/jupyter/jupyter-sphinx/issues/30#issuecomment-487914541 `SPHINXOPTS="-j8"` is mentioned but this immediately errors out.

Should this work or are we still waiting on upstream changes?

PR #39, opened 2019-05-01
Docs

### LINKED ISSUE #26: document handling packages that are not installed

### LINKED ISSUE #32: documentation
We'll probably only need one page, but it's the perfect place for putting an example.

- [ ] Update README
- [ ] Setup sphinx, add interesting examples both showing the output formats and package options.
- [ ] Deploy on RTD

jbweston:
Closes #32 and closes #26.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 # Jupyter Sphinx Extensions
 
-Jupyter-sphinx enables the rendering of Jupyter interactive widgets in sphinx
-documentation.
+``jupyter-sphinx`` enables running code embedded in Sphinx documentation and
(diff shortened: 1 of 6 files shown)


akhmerov:
Nice!

At a glance the only thing that's missing is a mention that this also automatically enables correct code highlighting.

PR #49, opened 2019-05-31
add javascript support

basnijholt:
With this we can (for example) include HoloViews, Bokeh, or plotly plots in the docs.

This is already used in https://adaptive.readthedocs.io

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -432,6 +432,12 @@ def cell_output_to_nodes(cell, data_priority, dir):
                     text=data,
                     rawsource=data,
                 ))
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
> I believe the main use for javascript output is pre-jupyterlab libraries injecting their javascript into the notebook after one calls something like init_notebook.

Yeah, it is unfortunate that some packages like bokeh use javascript injection instead of relying on notebook extensions.

Plotly plots should already work with jupyter-sphinx without javascript output since plotly moved to using jupyter interactive widgets.

JavaScript injection pauses other issues. I am fine with adding it here since we can assume that the documentation content is trusted however I don't have an opinion on the precedence order.

PR #52, opened 2019-06-11
show output to stderr

### LINKED ISSUE #48: handle stderr
Right now we are including `stdout` but ignoring stderr ([relevant lines](https://github.com/jupyter-widgets/jupyter-sphinx/blob/master/jupyter_sphinx/execute.py#L380-L387)).

We should also handle stderr on a similar footing, but not immediately sure what's the optimal approach.

jbweston:
Now we raise an exception if a cell prints to stderr. If the `:stderr:` option is provided on a cell then any output on stderr is printed as output and no exception is raised.

Closes #48.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -56,7 +56,7 @@ much richer output than mere text; plots, for example:
     from matplotlib import pyplot
     %matplotlib inline
 
-    x = np.linspace(0, 2 * np.pi)
+    x = np.linspace(1E-3, 2 * np.pi)
 
     pyplot.plot(x, np.sin(x) / x)
     pyplot.plot(x, np.cos(x))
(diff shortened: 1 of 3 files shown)


akhmerov:
I've [asked](https://groups.google.com/forum/#!topic/sphinx-users/VMpg4d6i1GI) about highlighting in sphinx-users mailing list.

akhmerov:
There doesn't seem to be a proper way to apply highlighting to stderr :(

I think an appropriate way to handle this is to provide a warning admonition in the docs where we discuss `stderr`.

jbweston:
@anton how about now (I amended the top commit to add the admonition in the docs)

akhmerov:
Yes, very good, thanks.

The CI failure is a [problem on Travis side](https://travis-ci.community/t/python-3-5-failing-on-download-in-bionic/3829/2), so I'm going to ignore it, hoping they get it back together.

PR #53, opened 2019-06-18
Add thebelab support

### LINKED ISSUE #41: support thebelab out of the box
This is more work than just flicking a switch: we need to change the behavior of hidden code from not being shown to being in DOM but invisible until thebelab is activated.

hugokerstens:
Closes #41 

This PR:
- Transforms JupyterCellNodes to thebelab cells and output for the html builder
- Adds a 'Make Live' button that bootstraps thebelab
- Adds some extra assets needed for thebelab


hugokerstens:
TODO:
- Thebelab currently does not work because the config is not correctly loaded
- Add a cell attribute to disable thebelab for that specific cell
- Better position for the 'Make Live' button

akhmerov:
Why do we add thebelab-latest instead of loading via CDN? Are we relying on unreleased features here?

hugokerstens:
> Why do we add thebelab-latest instead of loading via CDN? Are we relying on unreleased features here?

We are indeed relying on the predefined output functionality that is not yet released.

akhmerov:
I think that one is [already released](https://thebelab.readthedocs.io/en/latest/_static/demo-preview.html).

hugokerstens:
Never mind, I was confused with the other fix I implemented, which is not released. Let me load it from the CDN.

akhmerov:
providing thebelab config via conf.py doesn't seem to work right now

image

This is with the following `thebelab_config.js`:

```
{"requestKernel": true, "binderOptions": {"repo": "jupyter-widgets/jupyter-sphinx"}, "predefinedOutput": true}
```

akhmerov:
There are two problems here:

- Thebelab config is javascript and not JSON, so we may not write it using `dump`
- Even with a correct config, thebelab seems to not like loading it from an external file. Inlining the same script seems to work.

akhmerov:
I also found a couple of bugs with:

-  `copy_asset` vs `copy_asset_file` (the former accepts only directory, not a filename).
- `isinstance(thebe_config, file)`: `file` not defined.

I pushed my WIP to https://github.com/akhmerov/jupyter-sphinx/tree/thebelab-support

hugokerstens:
> providing thebelab config via conf.py doesn't seem to work right now

> Even with a correct config, thebelab seems to not like loading it from an external file. Inlining the same script seems to work.

That's right, it probably only supports an inline script. I'll edit that.

> Thebelab config is javascript and not JSON, so we may not write it using `dump`

JSON stands for 'JavaScript Object Notation', so it should be fine.

hugokerstens:
> I also found a couple of bugs with:
> 
> * `copy_asset` vs `copy_asset_file` (the former accepts only directory, not a filename).
> * `isinstance(thebe_config, file)`: `file` not defined.
> 
> I pushed my WIP to https://github.com/akhmerov/jupyter-sphinx/tree/thebelab-support

Whoops, I didn't test that code yet... Thanks for the fix, I'll add it to my branch.

akhmerov:
> JSON stands for 'JavaScript Object Notation', so it should be fine.

Indeed, you're right. Thebelab examples have some values unquoted, and I initially thought that is what causes the error. Turns out it was the inline thing, and everything works just fine.

akhmerov:
Here's a strategy for including the button:

- Decide on a reasonable place to include it within the page.
- Add a boolean flag for including it right there
- Explain how to include it at some other place, probably providing a directive or a tool for it.

jupyter_sphinx/execute.py
```diff
@@ -443,14 +503,43 @@ def cell_output_to_nodes(cell, data_priority, dir):
...
+def attach_outputs(output_nodes, node, thebe_config):
+    if thebe_config and not node.attributes['no_thebelab']:
+        code_class = 'thebelab-code'
+        if node.attributes['hide_code']:
+            code_class += ' thebelab-hidden'
+
+        code = node.astext()
+
+        node.children = [docutils.nodes.raw(
+            text='<pre class="{code_class}" data-executable="true" data-language="python">' \
```
jbweston: is it correct to hardcode Python as the language here?

jupyter_sphinx/execute.py
```diff
@@ -443,14 +503,43 @@ def cell_output_to_nodes(cell, data_priority, dir):
...
+                '{data}' \
```
jbweston: Also, how is this rendered for writers other than HTML (does it degrade nicely?)
Perhaps we should use specific node classes rather than relying on `nodes.raw`.

jupyter_sphinx/execute.py
```diff
@@ -443,14 +503,43 @@ def cell_output_to_nodes(cell, data_priority, dir):
...
+                '</pre>'.format(data=code, code_class=code_class),
+            format='html',
+        )]
+
+        if not node.attributes['hide_output']:
+            # We ignore the code_below attribute since this is not supported with thebelab
+            node.children.append(docutils.nodes.raw(
+                text='<div class="thebelab-output" data-output="true">',
+                format='html',
+            ))
```
jbweston: same as above; how does this degrade when not using an HTML writer?

jbweston: and shouldn't we add children to this `div` node, rather than having the opening and closing tags as separate nodes?

jupyter_sphinx/execute.py
```diff
@@ -546,6 +635,61 @@ def sphinx_abs_dir(env):
...
+        return
+
+    # Force config values to make thebelab work correctly
+    thebe_config['predefinedOutput'] = True
+    thebe_config['requestKernel'] = True
+
+    # Specify the thebelab config inline, a separate file is not supported
+    doctree.append(docutils.nodes.raw(
+        text=f'\n<script type="text/x-thebe-config">\n{json.dumps(thebe_config)}\n</script>',
+        format='html'
```
jbweston: Same as above: how does this look for non-HTML writers? I can't remember if there is a specific node type for javascript that is properly skipped for non-HTML writers.

akhmerov: Because format is specified, it will not appear in non-HTML writers, that's AFAIK how raw nodes work.

jupyter_sphinx/execute.py
akhmerov: We already have the language from the kernel, so we should use that.

akhmerov:
We should still add documentation and tests.

akhmerov:
After rebasing on #57 I obtained this output from `make -C doc html`:

[jupytersphinx.pdf](https://github.com/jupyter-widgets/jupyter-sphinx/files/3365995/jupytersphinx.pdf)

I observe the following limitations:

1. `:hide-code:` doesn't work: 
image
2. `:code-below:` doesn't work (also not in the html builds) image

Given how thebelab and sphinx work, these require some thinking on how to fix.

hugokerstens:
> Given how thebelab and sphinx work, these require some thinking on how to fix.

We could just place the code below the output. On pressing the button, reorder them, and then activate Thebelab. Allowing the code below the output while the Thebelab cell is active is quite difficult I suppose.

jbweston:
Remove usage of f-strings to maintain compatibility with Python 3.5

akhmerov:
> We could just place the code below the output. On pressing the button, reorder them, and then activate Thebelab.

Indeed, let's go with that if you think it's doable.

jupyter_sphinx/execute.py
hugokerstens: The language that is specified in the kernel is something like `ipython3`, while thebelab expects `python`. I do not see an easy way to not hardcode this.

akhmerov:
The render is now rather slow, here's a google report: https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fantonakhmerov.org%2Fmisc%2Fjupyter-sphinx%2F

In particular, eliminating render-blocking resources (js assets) seems to be relevant.

akhmerov:
@hugokerstens the commit message reads "partially support code-below", while everything seems to work. What's partial?

jupyter_sphinx/execute.py
akhmerov: This is only used for highlighting, right?

hugokerstens:
> @hugokerstens the commit message reads "partially support code-below", while everything seems to work. What's partial?

It is partially since the code is no longer below the source after activating thebelab on the page.

akhmerov:
> It is partially since the code is no longer below the source after activating thebelab on the page.

That's "fully" in my book.

jupyter_sphinx/execute.py
akhmerov: If it's only highlighting, then we should use the one from the kernel and open an issue on thebelab side instead.

hugokerstens:
The widgets do not work after activating thebelab on the page. They work again after you execute the code inside the cell. I suppose this has is caused by moving them around the DOM. IMO this is just a minor issue though, and certainly not an easy fix (if at all possible).

akhmerov:
> The widgets do not work after activating thebelab on the page. They work again after you execute the code inside the cell. I suppose this has is caused by moving them around the DOM. IMO this is just a minor issue though, and certainly not an easy fix (if at all possible).

Agreed, we'll open the follow-up issues after merging this.

jupyter_sphinx/execute.py
```diff
@@ -124,6 +131,8 @@ class JupyterCell(Directive):
         If provided, the cell output will not be displayed in the output.
     code-below : bool
         If provided, the code will be shown below the cell output.
+    no-thebelab : bool
```
akhmerov: This seems an unnecessary complication that will see little use, let's remove it and reinstate if there is a request.

akhmerov:
`:hide-code:` doesn't work now *without* thebelab activated.

akhmerov:
Other than those two last issues LGTM

hugokerstens:
> `:hide-code:` doesn't work now _without_ thebelab activated.

For me it does? Also the tests confirm that it works.

akhmerov:
> For me it does? Also the tests confirm that it works.

You're right, must be a side-effect of not running `make clean` before builds.

jupyter_sphinx/execute.py
hugokerstens: While this specific implementation is not optimal and would not get used a lot, there should at least be _some_ way to not enable thebelab everywhere if you want to use it only at specific places/pages.

akhmerov: Why should there be some way? Thebelab isn't visible before the user clicks the button, and so this doesn't seem to be a problem.

I suggest to wait for a use case because it's easy to implement this back, but having a specific application would allow us to make a more informed choice (maybe it makes more sense on the level of kernel) without running into backwards compatibility issues.

hugokerstens: I agree that it is a good idea to wait until we see a real life use case. Removed in 4cf2aec.

akhmerov:
:tada: :tada: :tada:

PR #68, opened 2019-07-16
include thebelab javascript and css in source distributions

### LINKED ISSUE #67: thebelab does not work when using installed jupyter-sphinx
the `thebelab-helper.js` and `thebelab.css` are not packaged with jupyter-sphinx, so thebelab does not work for documentation built using an installed version of jupyter-sphinx.

jbweston:
without these files thebelab does not work for documentation
built using installed versions of jupyter-sphinx

Closes #67.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include LICENSE
+recursive-include jupyter_sphinx/thebelab/*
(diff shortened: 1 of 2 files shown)


jbweston:
We should release 0.2.1 after this is merged, as this is a pretty big bug.

akhmerov:
Thanks for fixing this!

PR #69, opened 2019-07-16
make ThebelabSourceCell into a container with original code inside

### LINKED ISSUE #66: Code highlighting no longer works
check out: https://jupyter-sphinx.readthedocs.io/en/latest/

probably related to thebelab

akhmerov:
This fixes code highlighting before thebelab is activated and closes #66


akhmerov:
I reckoned we could go for other easy to close issues before releasing 0.2.1

akhmerov:
It's not that: before we were including *contents* of the child inside directly (so the source with no formatting), now we're letting the child literal block render itself.

pygments produces unreadable DOM, but thebelab uses the correct jQuery way of getting only the text.

PR #70, opened 2019-07-16
Bugfix/sphinx version

akhmerov:


jbweston:
I don't know when sphinx started using `contents.rst` as the root document, but: https://stackoverflow.com/questions/56336234/build-fail-sphinx-error-contents-rst-not-found

Seems we should add 
```
master_doc = 'index'
```
to our conf.py

PR #73, opened 2019-07-29
add possibility to disable raising exception and printing stderr

### LINKED ISSUE #72: Allow disabling of stderr printing
Sometimes when importing certain packages there are various warnings that are raised. These warnings are often unavoidable and are not the fault of the end user.

Currently we raise an exception whenever anything is printed to `stderr`, however it seems that the correct way to handle these warning messages is just to suppress them entirely; i.e. not to print to `stderr` and also not to raise an exception.

I propose a configuration parameter that suppresses `stderr` and does not raise an exception when it is printed to.

jbweston:
Closes #72.

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -379,8 +379,8 @@ class ExecuteJupyterCells(SphinxTransform):
                           if output['output_type'] == 'stream'
                              and output['name'] == 'stderr']
                 if stderr and not node.attributes['stderr']:
-                    raise ExtensionError('Cell printed to stderr:\n{}'
-                                         .format(stderr[0]['text']))
+                    logger.warning('Cell printed to stderr:\n{}'
+                                   .format(stderr[0]['text']))
 
             try:
                 lexer = notebook.metadata.language_info.pygments_lexer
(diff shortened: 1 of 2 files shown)


akhmerov:
Thanks for finding the issue and proposing the fix. The implementation looks solid, however the main choice that we have to make is the UI for controlling this behavior.

We have to choose between default hiding, default showing, default raising on stderr.

Further, we should decide at which levels out of conf.py / single kernel / single cell should the user be able to control this behavior.

This decision depends on the typical usage patterns that we anticipate.

akhmerov:
Also I think that emitting a sphinx warning if there's stderr is a useful behavior.

akhmerov:
Possible usage scenarios, sorted according to my estimation for how common they are, from most to least frequent.

- The user does not want to show stderr but wants to know that the code produces stderr/warnings throughout the complete documentation
- The user expects the code to produce no stderr, considers stderr an error
- The user has a handful cases where they want to actually show stderr

akhmerov:
Assuming the above collection of scenarios is correct, it seems that the correct approach is:

- To always emit extension warnings when stderr is produced. The users may then use sphinx tools to either convert these to errors or silence them
- To allow a per-cell option for including stderr, which would then produce no warning

akhmerov:
@jbweston what do you think?

jbweston:
Yes that seems like the correct thing

akhmerov:
What should we call the new option? Is just `:stderr:` clear enough?

jbweston:
I think so; it's not a new option; it just does what it did before. AFAICT the only difference now is that there is a warning raised and not an error

akhmerov:
Thanks for implementing the changes. Will you have a chance to fix the tests as well?

jbweston:
That's everything fixed

akhmerov:
Thanks!

PR #81, opened 2019-10-02
added linenos (line numbers) flag, updated docs, and basic tests

seanpue:
I added a ``:linenos:`` flag to add line numbers to the source code. One aim here is to distinguish code and output.

It looks a little weird in the Alabaster theme, as it cuts the lines length, but it looks pretty nice in readthedocs.  It generates a standard html table, so it could be CSS styled as one pleases.

It also shows up in LatexPDF.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.tox/
 *.py[cod]
 
 # Packages
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -175,6 +175,24 @@ produces:
 
     print('this code is below the output')
 
+You may also add *line numbers* to the source code with ``:linenos:``::
+
+    .. jupyter-execute::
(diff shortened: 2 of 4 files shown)


akhmerov:
AFAIK `:linenos:` is the name used in sphinx for `literal-block`, so this seems appropriate.

seanpue:
Hi @jbweston I agree it's a bit weird sounding, but it's from Sphinx ``code-block``: https://www.sphinx-doc.org/en/1.5/markup/code.html#line-numbers . There is also  ``:emphasize-lines``: and ``:lineno-start:`` that take values. Those might be nice, too. I'm not sure if those will work on the ``literal-block`` in the container, however. Also, @akhmerov should it tie into Thebes, as well, or is that not relevant? I have not experimented with it yet.

PR #82, opened 2019-10-04
added continue_linenos directive, doc updates, and tests

seanpue:
This adds a :continue_linenos: directive that allows you to continue the previous cell's line numbers, i.e. if it is flagged with `linenos` or `continue_linenos`.

Any cells in between with the `hide_code` directive will not reset the line numbering.
If there is no previous line-numbered cell, it starts the line numbering at 1. 

I am finding it useful in documentation to show sections of related code, e.g. using the same variables, as restarting numbering from 1 in those cases looks odd.

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -154,7 +154,7 @@ class JupyterCell(Directive):
         'hide-output': directives.flag,
         'code-below': directives.flag,
         'linenos': directives.flag,
-        'continue_linenos': directives.flag,
+        'continue-linenos': directives.flag,
         'raises': csv_option,
         'stderr': directives.flag,
     }
@@ -190,7 +190,7 @@ class JupyterCell(Directive):
             hide_output=('hide-output' in self.options),
             code_below=('code-below' in self.options),
             linenos=('linenos' in self.options),
-            continue_linenos=('continue_linenos' in self.options),
+            continue_linenos=('continue-linenos' in self.options),
             raises=self.options.get('raises'),
             stderr=('stderr' in self.options),
         )]
@@ -398,7 +398,7 @@ class ExecuteJupyterCells(SphinxTransform):
 
             # Add line numbers to code cells if linenos directive is set.
             # Continue line numbers from previous cell with line numbers
-            # if continue_linenos directive is set.
+            # if continue-linenos directive is set.
             linenostart = 1
             for node in nodes:
                 source = node.children[0]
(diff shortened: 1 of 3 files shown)


akhmerov:
I'm hesitant to introduce too many directive options; do you think this can be reasonably replaced with a global or a kernel-wide configuration option instead?

seanpue:
There could be a global option to number the lines, but I am trying to use the line numbering semantically, to show different related lines of code, as a copyable, more readable and less annoying version of `ipythontb`. Here is a documentation link:  https://graphtransliterator.readthedocs.io/en/latest/usage.html#configuration

I also modified stderr and the CSS to try and make it more readable. Somehow, setting the output of stderr as a literal_block with the `error` class and setting `rawsource=None` and using `text` outputs like Jupyter warnings. I also want to have the option to limit the length of the traceback messages, probably by modifying the text directly.

jupyter_sphinx/execute.py
```diff
@@ -394,11 +396,23 @@ def apply(self):
...
+                source = node.children[0]
+                if node["continue_linenos"]:
+                    source["linenos"] = True
+                    source["highlight_args"] = {'linenostart': linenostart}
+                elif node["linenos"]:
+                    linenostart = 1
                     source["linenos"] = True
+                # Advance linenostart or reset it
+                if node["continue_linenos"] or node["linenos"]:
+                    linenostart += len(source.rawsource.split("\n"))
```
jbweston: maybe
```suggestion
                    linenostart += source.rawsource.count("\n") + 1
```

jbweston:
> I also modified stderr and the CSS to try and make it more readable.

I don't see such a modification in this PR.

jbweston:
> I'm hesitant to introduce too many directive options; do you think this can be reasonably replaced with a global or a kernel-wide configuration option instead?

Why do you consider directive options more unwieldy than configuration options?

akhmerov:
> Why do you consider directive options more unwieldy than configuration options?

I think that having to provide options with *every* code snippet is inconvenient since it increases the amount of boilerplate. The typical use case shouldn't require providing an option to most directives.

seanpue:
That would work. Since the linenos directive is there now, it could be used to restart line numbering in a documentation section. That would certainly work in my use case. Or do you have another solution?

Another way to get rid of the boilerplate would be to have default directives as a config option.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jupyter_sphinx/execute.py
seanpue: Much nicer, thanks!

seanpue:
I added a new version, #83 using config options. Let me know what you think. I will update the tests for it.

seanpue:
Closed in favor of #83.

PR #83, opened 2019-10-06
Continue linenos through setting

seanpue:
Alternative to #82 . Adds `jupyter_sphinx_linenos` and `jupyter_sphinx_continue_linenos` settings. Both will turn on line numbering by default (equivalent to directive `:linenos:`). `jupyter_sphinx_continue_linenos` will continue the line number from the previous cell by default. The `linenos` directive can be used to restart line numbering at one. Updated docs.

Tests are insufficient.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -193,6 +193,31 @@ produces:
     print('B')
     print('C')
 
+You may also *continue line numbers* from the previous cell with ``:continue_linenos:``::
+
+  .. jupyter-execute::
+      :linenos:
+
+      print('A')
+
+  .. jupyter-execute::
+      :continue_linenos:
+
+      print('B')
+      print('C')
+
+produces:
+
+.. jupyter-execute::
+    :linenos:
+
+    print('A')
+
+.. jupyter-execute::
+    :continue_linenos:
+
+    print('B')
+    print('C')
 
 Controlling exceptions
 ----------------------
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -330,11 +330,9 @@ jupyter_execute_kwargs
 
 jupyter_sphinx_linenos
 
-    Boolean argument it use line numbering in code cell display by default. Equivalent
-    to setting the `linenos` directive for all code cells.
+    Whether to show line numbering in all ``jupyter-execute`` sources.
 
 jupyter_sphinx_continue_linenos
 
-    Boolean argument to use line numbering in code cells and continue line numbering
-    from the previous cell. When set, line number can be reset to start at one by using
-    the directive `linenos`.
+    Whether to show continuous line numbering in all ``jupyter-execute`` sources.
+    Line numbering can be reset to one with a ``linenos`` directive.
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -335,4 +335,3 @@ jupyter_sphinx_linenos
 jupyter_sphinx_continue_linenos
 
     Whether to show continuous line numbering in all ``jupyter-execute`` sources.
-    Line numbering can be reset to one with a ``linenos`` directive.
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -175,6 +175,24 @@ produces:
 
     print('this code is below the output')
 
+You may also add *line numbers* to the source code with ``:linenos:``::
+
+    .. jupyter-execute::
+        :linenos:
+
+        print('A')
+        print('B')
+        print('C')
+
+produces:
+
+.. jupyter-execute::
+    :linenos:
+
+    print('A')
+    print('B')
+    print('C')
+
 
 Controlling exceptions
 ----------------------
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -193,7 +193,7 @@ produces:
     print('B')
     print('C')
 
-You may also *continue line numbers* from the previous cell with ``:continue_linenos:``::
+You may also *continue line numbers* from the previous cell with ``:continue-linenos:``::
 
   .. jupyter-execute::
       :linenos:
@@ -201,7 +201,7 @@ You may also *continue line numbers* from the previous cell with ``:continue_lin
       print('A')
 
   .. jupyter-execute::
-      :continue_linenos:
+      :continue-linenos:
 
       print('B')
       print('C')
@@ -214,7 +214,7 @@ produces:
     print('A')
 
 .. jupyter-execute::
-    :continue_linenos:
+    :continue-linenos:
 
     print('B')
     print('C')
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -193,32 +193,6 @@ produces:
     print('B')
     print('C')
 
-You may also *continue line numbers* from the previous cell with ``:continue-linenos:``::
-
-  .. jupyter-execute::
-      :linenos:
-
-      print('A')
-
-  .. jupyter-execute::
-      :continue-linenos:
-
-      print('B')
-      print('C')
-
-produces:
-
-.. jupyter-execute::
-    :linenos:
-
-    print('A')
-
-.. jupyter-execute::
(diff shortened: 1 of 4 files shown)


seanpue:
Tests are done. That was easier than I thought it would be! Let me know what you think of this version.

akhmerov:
Yes, I think this is a better implementation, and I propose to go ahead with this version.

doc/source/index.rst
```diff
@@ -328,3 +327,14 @@ jupyter_execute_kwargs
 
     Keyword arguments to pass to ``nbconvert.preprocessors.execute.executenb``, which controls how
     code cells are executed. The default is ``dict(timeout=-1, allow_errors=True)``.
+
+jupyter_sphinx_linenos
+
+    Boolean argument it use line numbering in code cell display by default. Equivalent
```
akhmerov: This sentence seems to have a typo.

akhmerov: I think saying that it's an argument is redundant. How about the following clarification:

"Whether to show line numbering in all ``jupyter-execute`` sources."

doc/source/index.rst
```diff
@@ -328,3 +327,14 @@ jupyter_execute_kwargs
...
+    to setting the `linenos` directive for all code cells.
```
akhmerov: It might not be obvious to the user what we call "code cells", I think we should avoid calling the directives "code cells" unless we clearly introduce the concept.

seanpue: Ok, thanks, that is much clearer. I have adjusted the docs. Please check those out. I second going ahead with this version.

seanpue:
Ok, thanks, your suggestions  are much clearer. I have adjusted the docs. Please check those out. I second going ahead with this version.

jupyter_sphinx/execute.py
```diff
@@ -394,11 +399,24 @@ def apply(self):
...
+            # Update current line numbers from cell if jupyter_sphinx_continue_linenos
+            # is set.
+            linenostart = 1
             for node in nodes:
+                source = node.children[0]
+                if linenos_config or continue_linenos:
+                    source["linenos"] = True
                 if node["linenos"]:
-                    source = node.children[0]
+                    linenostart = 1
```
akhmerov: I think this would be a confusing interaction: it's best if the option `continue_linenos` works the same regardless of whether line numbers are turned on globally or not.

jupyter_sphinx/execute.py
```diff
@@ -189,6 +192,7 @@ def run(self):
             hide_output=('hide-output' in self.options),
             code_below=('code-below' in self.options),
             linenos=('linenos' in self.options),
+            # continue_linenos=('continue-linenos' in self.options),
```
akhmerov: Please remove commented out code.

jupyter_sphinx/execute.py
```diff
@@ -327,7 +331,8 @@ def apply(self):
         default_kernel = self.config.jupyter_execute_default_kernel
         default_names = default_notebook_names(docname)
         thebe_config = self.config.jupyter_sphinx_thebelab_config
-
+        linenos_config = self.config.jupyter_sphinx_linenos
```
akhmerov: As I also wrote in the other place, I think we shouldn't distinguish what was the source of turning line numbers on.

jupyter_sphinx/execute.py
seanpue: Done.

jupyter_sphinx/execute.py
seanpue: This just reads the config option. I could rename it `auto_linenos` if that is clearer. As I note below, I am not necessarily following your point, sorry.

jupyter_sphinx/execute.py
seanpue: Not sure if I am following you. This allows restart from one if :linenos: directive is added to a cell and `jupyter_sphinx_continue_linenos` config option is set. It does not matter if `linenos` is turned on as a directive or by the `jupyter_sphinx_linenos` global option.

akhmerov: I'm saying that a document with every input containing `:linenos:` should always render exactly the same as if the config option `jupyter_sphinx_linenos` was set.

If I'm not mistaken, right now if the user sets `jupyter_sphinx_continue_linenos`, and all cells contain `:linenos:` option, the line numbers would be reset in each cell, resulting in a different output.

seanpue: Ok. That makes sense. I took that out. I think it would be nice to be able to reset the numbering, however. Would a :linenos-reset: directive be okay?  I don't there if there is a way to do "magic" that would accomplish the same thing. It could be ":_linenos-reset".  :)

tests/test_execute.py
```diff
@@ -141,6 +141,49 @@ def test_linenos(doctree):
...
+    cell0, cell1, cell2 = tree.traverse(JupyterCellNode)
+    assert cell0.children[0].attributes['linenos']
+    assert cell0.children[0].attributes['highlight_args']['linenostart'] == 1
+    assert cell0.children[0].rawsource.strip() == "2 + 2"
+    assert cell0.children[1].rawsource.strip() == "4"
+
+    assert cell1.children[0].attributes['linenos']
+    assert cell1.children[0].attributes['highlight_args']['linenostart'] == 2
+    assert cell1.children[0].rawsource.strip() == "3 + 3"
+    assert cell1.children[1].rawsource.strip() == "6"
```
jbweston: I see that we don't use `cell2` here; shall we remove it?

jbweston:
@akhmerov your replies to @seanpue's responses would be good so we can merge this

jupyter_sphinx/execute.py
akhmerov: I feel that any extra setting would be too niche. Would it be possible to achieve a similar result by resetting the count on any hidden input?

jupyter_sphinx/execute.py
akhmerov: I see, looks OK.

akhmerov:
@seanpue I'd like to release a new version (containing your awesome styling solution). Would you like to include this as well? Alternatively we can let it hang, since the main problem with styling is now fully addressed.

tests/test_execute.py
seanpue: Removed it.

seanpue:
@akhmerov Great! I think this is now ready to be included, as well.

jupyter_sphinx/execute.py
```diff
@@ -394,11 +397,19 @@ def apply(self):
...
+            # jupyter_sphinx_continue_linenos are set in the configuration,
+            # or the linenos directive is set.
+            # Update current line numbers from cell if jupyter_sphinx_continue_linenos
+            # is set.
+            linenostart = 1
             for node in nodes:
-                if node["linenos"]:
-                    source = node.children[0]
+                source = node.children[0]
+                if linenos_config or continue_linenos or node["linenos"]:
```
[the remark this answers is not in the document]
seanpue: Yeah, that just turns on line numbering if `jupyter_sphinx_continue_linenos` is set in `conf.py`.

jupyter_sphinx/execute.py
seanpue: Okay, let's hold off on that. This can go merge as is without it.

SeanDS: For what it's worth, I wanted exactly this reset feature as soon as I started using the `continue_linenos` `conf.py` setting. I'm disappointed it was taken out.

seanpue: Another way of doing a restart of the numbering restart would be to add support for the Sphinx code-block directive`` lineno-start``, which could be set to 1 if continuation is on. I think that would be generally useful, e.g. for referencing earlier code, and intuitive to use. https://www.sphinx-doc.org/en/1.5/markup/code.html#line-numbers

PR #85, opened 2019-10-07
Adds jupyter_sphinx_truncate_traceback option. Fixes #84.

### LINKED ISSUE #84: config option to truncate traceback
It would be useful to be able to be able to have the option to truncate the display of error traceback. When writing documentation, it is useful just to show that an error occurred without showing the full traceback. This practice is commonly done in terminal view.

There is ``sys.tracebacklimit`` setting. However, it does not work in Jupyter and needs a rather complex override: https://stackoverflow.com/questions/46222753/how-do-i-suppress-tracebacks-in-jupyter

I would propose adding a ``jupyter_sphinx_truncate_traceback`` option that would just show the first two lines of the normal traceback message followed by an ellipsis, e.g.
```
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
...
```

seanpue:
Adds jupyter_sphinx_truncate_traceback option that truncates traceback to first two lines followed by an ellipsis, e.g.

```
---------------------------------------------------------------------------
ZeroDivisionError                         Traceback (most recent call last)
...
```

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -328,3 +328,8 @@ jupyter_execute_kwargs
 
     Keyword arguments to pass to ``nbconvert.preprocessors.execute.executenb``, which controls how
     code cells are executed. The default is ``dict(timeout=-1, allow_errors=True)``.
+
+jupyter_sphinx_truncate_traceback
+
+    Whether to truncate trackback on a error, displaying only the first two lines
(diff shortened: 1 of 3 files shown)


seanpue:
Closed following suggestion #84.

Issue #71, opened 2019-07-21
explore how to improve styling


akhmerov:
We have two problems:

- Right now the inputs and outputs are not visibly different
- stderr isn't different from stdout

We should try to improve on both of these, but I think it's rather hard.

danielballan:
I really like this library. Thank you for contributing this work to the Jupyter ecosystem!

As a longtime user of the IPython sphinx extension, it occurs to me that one way to make inputs and outputs visibly different is to display `In [...]` and `Out [...]` prompts to the left of the code blocks. Unlike the IPython sphinx extension, however, the prompts would apply to a whole code block at once, not to individual lines. (See [this page from the pandas documentation](https://pandas.pydata.org/pandas-docs/stable/getting_started/10min.html) for exampel of what prompts-per-line look like.) This makes the code examples more compact and also easier to copy/paste.

Do you think this is a suitable approach?

akhmerov:
@danielballan thanks for the idea. It's an possible choice, but there are some subtleties to consider:

- jupyter distinguishes the *output* and the display data. Within the notebook/lab only output is labeled with `Out[...]`:
![image](https://user-images.githubusercontent.com/2069677/62621926-60c19400-b91d-11e9-97b7-d608e0d0989f.png)
- Documentation may reasonably hide some cells, for example boilerplate with imports. It will then be confusing to the readers why the first code they see is `In[3]`.
- The layout isn't obvious if we're displaying rich output. If we just show text, then the formatting is clear, but what kind of DOM should we produce if we have an image?
- What to do if the input or the output are hidden?

pohlt:
I just recently realized that both the code fragments and their input are styled as code. While this is an obvious choice for the code fragments, it's kind of misleading for the output. When I do a `print("The world is nice")`, the "is" in the output is emphasized as a keyword. The output of pre-formatted text is also messed up (e.g. when printing a numpy array).

Again, I don't know the internals, so what I'm proposing might not be feasible, but how about styling the output as pre-formatted text (and optionally giving it a different background color and/or font). IIRC, pre-formatted texts are part of reStructuredText and Sphinx.

akhmerov:
OK, highlighting is outright a bug; that we should fix. Styling should be within reach, however I don't know off the top of my head how to make it consistent across different formats (latex / html).

akhmerov:
Done; the current master should have no output highlighting (which was there due to https://github.com/sphinx-doc/sphinx/issues/4679).

[...]

PR #86, opened 2019-10-08
Stab at adding stderr differentiation

seanpue:
This attempts to add a separate output style for `stderr`, addressing #71. 

It adds a `literal_block` without highlighting (by not settings `rawsource` in `literal_block()`, which disables Pygment), inside a `container`. Both are set to classes `error` and `stderr`. `Error` is common in Sphinx themes, especially as a `div` (container) style, whereas `stderr`, applied after, is not, and will allow CSS customization by the user for this particular output.

The HTML result is the following:
```
<div class="error stderr docutils container">
<pre class="error stderr literal-block">{{ ERROR MESSAGE}}
</pre>
</div>
```

LatexPDF results are differentiated by the absence of borders. Tested in Alabaster and sphinx_rtd_theme.

### DIFF
--- a/doc/source/_static/custom.css
+++ b/doc/source/_static/custom.css
@@ -1,0 +1,7 @@
+/*
+Custom CSS for stderr stream output.
+*/
+
+.stderr {
+    background-color: #FCC;
+}
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -41,3 +41,4 @@ jupyter_sphinx_thebelab_config = {
         "repo": "jupyter/jupyter-sphinx",
     },
 }
+html_static_path = ["_static/custom.css"]  # css overrides for Alabaster theme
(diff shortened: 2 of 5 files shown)


jbweston:
Nice one! Can you post a screenshot of what this looks like with the standard sphinx theme?

seanpue:
Here it is without any CSS adjustments in Alabaster. (Alabaster has a background color defined for `pre`, so the text background is showing up gray).
Screen Shot 2019-10-08 at 3 06 52 PM

Here it is with a custom.css changing background color for `stderr`:
```
.stderr {
    background-color: #FCC
}
```
Screen Shot 2019-10-08 at 3 20 31 PM

And here is the PDF output:

Screen Shot 2019-10-08 at 3 22 32 PM

akhmerov:
Nice!

I think the only thing remaining to do here is updating the docs to remove the warning about the lack of styling.

seanpue:
Great! I removed the warning, added a note with a hyperlink hint about how to adjust the CSS styling in the default theme, added a `custom.css` to fix the background color in `doc/source/_static`, and adjusted `conf.py` to load it.

jbweston:
Closes #71

Issue #91, opened 2019-11-12
Odd line spacing before/after jupyter-execute blocks with sphinx_rtd_theme


Rockgecko:
This is a really nice extension and looks like it will make it much faster to add example code to our docs. Thank you.

One minor issue is that I am getting a very big gap between lines of text and the next code block, and a much smaller gap after the code block before the next line of text (see image). This makes it harder to read as it isn't so clear that the text relates to the following block of code rather than the previous one (which it is closer to). 

I've added a custom style sheet to try and adjust this, but I can't work out which parameter controls this spacing. I've tried almost all the padding and margin options from the elements in the default jupyter-sphinx css file, but nothing seems to change it. I have no experience with css so might be missing something obvious. Can anyone provide guidance as to what to change? 

![image](https://user-images.githubusercontent.com/10621533/68691289-6ad1df00-056b-11ea-9330-453482571a26.png)

Many thanks!

jbweston:
Thanks for the report!

I'll take a look tomorrow

seanpue:
@Rockgecko This custom CSS should fix it for RTD:

  ``
  div.jupyter_container + p {
    margin-top: 24px;
  }
  ``

You can add it to a ``custom.css`` in ``_static`` and then add ``html_css_files=['custom.css']`` to ``conf.py``.

RTD (unlike Alabaster) adds the custom.css before jupyter-sphinx.css, so the custom.css requires the more specific div + p combination rather than an override of the div.jupyter_container margin.

@jbweston I am not sure what an elegant solution to this issue would be, as I do not think it can be addressed by adjusting jupyter-sphinx.css and still display the alabaster theme correctly.

Rockgecko:
@seanpue Great, that works and looks much better. Thank you for your help!

seanpue:
Good. @jbweston I found a solution to identify RTD in CSS and override the jupyter_container margin. 

@Rockgecko That custom CSS should not cause a problem once that is merged, but you can delete it once it is.

akhmerov:
Closing via #92.

PR #92, opened 2019-11-14
Fix sphinx_rtd_theme spacing

seanpue:
Fix for #91. Identifies sphinx_rtd_theme in CSS via the .rst-content div and overrides margins of jupyter_container to match sphix_rtd_theme spacing.


akhmerov:
Amazing, thanks a lot for the fix!

PR #95, opened 2020-01-04
Add emphasize-lines option

espdev:
This PR adds `:emphasize-lines:` option to `sphinx-execute` directive. This option allows you to highlight specified lines in source code just like in `code-block` [directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive:option-code-block-emphasize-lines).

For example:
```
.. jupyter-execute::
    :linenos:
    :emphasize-lines: 2,5-6

    d = {
        'a': 1,
        'b': 2,
        'c': 3,
        'd': 4,
        'e': 5,
    }
```
produces:

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -2,11 +2,9 @@
 
 import os
 from itertools import groupby, count
-from operator import itemgetter
 import json
 
-import sphinx
-from sphinx.util import logging
+from sphinx.util import logging, parselinenos
 from sphinx.util.fileutil import copy_asset
 from sphinx.transforms import SphinxTransform
 from sphinx.errors import ExtensionError
@@ -26,7 +24,6 @@ from jupyter_client.kernelspec import get_kernel_spec, NoSuchKernel
 
 import nbformat
 
-from ipywidgets import Widget
 import ipywidgets.embed
 
 from ._version import __version__
@@ -136,6 +133,8 @@ class JupyterCell(Directive):
         If provided, the code will be shown below the cell output.
     linenos : bool
         If provided, the code will be shown with line numbering.
+    emphasize-lines : comma separated list of line numbers
+        If provided, the specified lines will be highlighted.
     raises : comma separated list of exception types
         If provided, a comma-separated list of exception type names that
         the cell may raise. If one of the listed execption types is raised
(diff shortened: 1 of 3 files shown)


jupyter_sphinx/execute.py
```diff
@@ -185,6 +187,21 @@ def run(self):
...
+        if emphasize_linespec:
+            try:
+                nlines = len(content)
+                hl_lines = parselinenos(emphasize_linespec, nlines)
+                if any(i >= nlines for i in hl_lines):
+                    logger.warning(
+                        'Line number spec is out of range(1-{}): {}'.format(
+                            nlines, emphasize_linespec), location=location)
+                hl_lines = [i + 1 for i in hl_lines if i < nlines]
+            except ValueError as err:
```
akhmerov: I don't quite understand which error we're catching here, can you explain?

Or is this implementation a direct copy of what's in code-block?

akhmerov:
Thanks for the contribution!

Overall I am a bit hesitant to add all the options from the code block to `jupyter-execute` due to its growing complexity, however I would be happy to merge you need this functionality.

jupyter_sphinx/execute.py
espdev: > Or is this implementation a direct copy of what's in code-block?

It is about [the same](https://github.com/sphinx-doc/sphinx/blob/0319faf8f1503453b6ce19020819a8cf44e39f13/sphinx/directives/code.py#L134-L148). :)

We are catching ValueError from `parselinenos` function `invalid line number spec`.

akhmerov: Thanks, can you please refer to the original code and add this comment inline?

Other than that, everything is good to go as far as I can see.

espdev:
@akhmerov 

Yes, I do not like as well copying the existing code and options, but I am not sure we can re-use the code from `code-block` directive in `jupyter-execute`. 
However, sometimes we need to highlight the particular lines of the code to draw users' attention to these lines. I think it would be great to have this functionality in jupyter-sphinx.

jupyter_sphinx/execute.py
espdev: Done

akhmerov:
Thanks! All good to go.

espdev:
Thanks for merging!

PR #97, opened 2020-01-31
Fix rule that gets applied on widgets

### LINKED ISSUE #96: CSS rule applied on wrong elements
The following CSS rule:
```css
div.jupyter_container * {
    padding: 0;
    margin: 0;
}
```

is applied on the `LeafletJS` popups and controls on the following pages:
- https://ipyleaflet.readthedocs.io/en/latest/api_reference/popup.html (click on the marker to see the Popup)
- https://ipyleaflet.readthedocs.io/en/latest/api_reference/layers_control.html (Click on the button on top right of the Map to see it)

It breaks their visual aspect. As a comparison, here is what the popup should look like:
popup

I don't know what this rule should be instead, maybe:
```css
// Without the star
div.jupyter_container {
    padding: 0;
    margin: 0;
}
```
is enough.

martinRenou:
This will fix #96. Although I am not sure you would agree with the result.

### DIFF
--- a/jupyter_sphinx/css/jupyter-sphinx.css
+++ b/jupyter_sphinx/css/jupyter-sphinx.css
@@ -57,7 +57,7 @@ div.jupyter_container {
 div.jupyter_container div.highlight {
   background-color: #f7f7f7; /* for haiku */
 }
-div.jupyter_container * {
+div.jupyter_container {
     padding: 0;
     margin: 0;
 }


akhmerov:
@martinRenou thanks a lot for proposing a fix! I'm sorry, I'm not as well-versed in CSS as I'd like to be. Can you please explain why I would potentially not agree with the result?

Also CC-ing @seanpue, who implemented the current CSS, for an opinion.

martinRenou:
Because, although it fixes my Popup issue, it does not give the exact same visual aspect as it was before:

Before:
before

After: (There is a small gap between the code and the widget view, the right and left padding around the text changed)
after

akhmerov:
Not using a wildcard selector makes a lot of sense because otherwise it would be hard to ensure no css collisions with widgets.

If we want to improve and fine-tune the appearance, we should do it at a later stage using more specific css selectors.

martinRenou:
Thanks!

seanpue:
Sorry, I have been swamped. LGTM. @akhmerov, let me know if anything gets distorted, and I can add more specific CSS selectors.

PR #103, opened 2020-02-29
refactoring into submodules

choldgraf:
This refactors jupyter_sphinx into a few submodules:

* `ast.py`
* `execute.py`
* `thebelab.py`
* `utils.py`

and puts the sphinx `setup()` function in `__init__.py`.

the docs build and the tests pass for me. I still want to clean up the imports, and apply a `black` pass, but wanted to see if this looks reasonable to others first. If folks are +1 on this, then I'll add another commit that cleans up the import statements, and applies "black" and ensures a flake8 test passes.

The one user-facing change is that the extension will now be loaded with `jupyter_sphinx` instead of `jupyter_sphinx.execute`. Perhaps there's a way we could import one into the other to temporarily keep both behaviors?


akhmerov:
Wouldn't an `execute` module that's a complete copy of `__init__` (perhaps with a deprecation warning) not solve it?

Otherwise backwards compatibility breaking would also be OK, but then I'd like to go sure that also the directives and the AST are stabilized before making a release.

choldgraf:
ah good call - I just added a commit that is a little bit hacky, but I think accomplishes the same thing. Also added a deprecation warning. WDYT?

I'm not sure how to test *both* the old and new way...

choldgraf:
I just cleaned up the imports a bit (so they are more well-ordered) and ran `black` on the codebase. @akhmerov what do you think?

jupyter_sphinx/__init__.py
```diff
@@ -1 +1,256 @@
...
+    app.add_config_value("jupyter_sphinx_thebelab_config", None, "html")
+
+    app.add_config_value("jupyter_sphinx_thebelab_url", THEBELAB_URL_DEFAULT, "html")
+
+    # linenos config
+    app.add_config_value("jupyter_sphinx_linenos", False, "env")
+    app.add_config_value("jupyter_sphinx_continue_linenos", False, "env")
+
+    # Used for nodes that do not need to be rendered
+    def skip(self, node):
```
akhmerov: Here and later: wouldn't it better to factor the inline functions and constants out of `_setup`?

jupyter_sphinx/__init__.py
```diff
@@ -1 +1,256 @@
...
+    app.add_lexer("ipythontb", IPythonTracebackLexer())
+    app.add_lexer("ipython", IPython3Lexer())
+
+    app.connect("builder-inited", builder_inited)
+    app.connect("build-finished", build_finished)
+
+    return {"version": __version__, "parallel_read_safe": True}
+
+
+def setup(app):
```
akhmerov: This seems to be identical to `setup = _setup`; are you sure that we cannot just call `__init__.setup` from `execute`?

jupyter_sphinx/__init__.py
choldgraf: good point - latest push has this in it

jupyter_sphinx/__init__.py
choldgraf: ah yeah - I just refactored a bit more so something like this is done

jupyter_sphinx/execute.py
```diff
@@ -87,16 +59,16 @@ class JupyterKernel(Directive):
...
+        return [
+            JupyterKernelNode(
+                "",
+                kernel_name=self.arguments[0].strip() if self.arguments else "",
+                kernel_id=self.options.get("id", "").strip(),
+            )
+        ]
 
 
 class JupyterKernelNode(docutils.nodes.Element):
```
akhmerov: Shouldn't the nodes be in `ast`?

choldgraf: I was torn on this one - on one hand they're clearly AST objects, on the other they are explicitly about executing things. Do you have a preference? I'm happy to go with either

akhmerov: I don't have a strong preference neither, but I think like putting all nodes (including this one) would make codebase a bit easier to navigate. Therefore I suggest indeed migrating this to `ast`. It's not a big point, and I'm happy with it either way.

jupyter_sphinx/__init__.py
akhmerov: Thanks!

jupyter_sphinx/execute.py
choldgraf: done in the latest push!

jupyter_sphinx/ast.py
```diff
@@ -0,0 +1,351 @@
...
+import docutils
+from docutils.parsers.rst import Directive, directives
+from docutils.nodes import math_block
+from sphinx.util import parselinenos
+from sphinx.addnodes import download_reference
+
+import ipywidgets.embed
+import nbconvert
+
+import jupyter_sphinx as js
```
akhmerov: It's a cosmetic pont, but I prefer to not introduce abbreviations like this, except for the very standard `np` or `plt`.

At the same time, seeing where this is used, it seems that

```suggestion
from jupyter_sphinx import logger
```

would work even better.

choldgraf: ah - the reason I was doing this is because the logger gets initialized by Sphinx so it's not importable at first. I'll lazily import the logger as you show above but inside the run function

akhmerov: Ah, thanks for the explanation.

akhmerov:
Thanks a lot!

choldgraf:
woo - OK I'll try and upstream some more stuff this week

PR #107, opened 2020-03-10
Using a post-transform for the AST

choldgraf:
This PR implements the logic that myst-nb uses for handling outputs. It roughly does the following:

* Add a few more cell node types (CellInputNode, CellOutputNode, and CellOutputBundleNode)
* Wraps cells in these node types
* Execute will now add a CellOutputBundleNode with the outputs inserted into it
* a post-transform will now transform these into CellOutputNode, followed by another imagecollector step.

## Todo

Need to figure out how to get the testing working. Currently I believe it assumes that the code is executed at parse time, as opposed to in a post-transform. Anybody have an idea for how to do this? My sphinx+pytest skills are not great

### DIFF
--- a/jupyter_sphinx/__init__.py
+++ b/jupyter_sphinx/__init__.py
@@ -6,16 +6,21 @@ import docutils
 import ipywidgets
 import os
 from sphinx.util.fileutil import copy_asset
+from sphinx.errors import ExtensionError
 from IPython.lib.lexers import IPythonTracebackLexer, IPython3Lexer
 
 from .ast import (
     JupyterCell,
     JupyterCellNode,
+    CellInputNode,
+    CellOutputNode,
+    CellOutputBundleNode,
     JupyterKernelNode,
     JupyterWidgetViewNode,
     JupyterWidgetStateNode,
     WIDGET_VIEW_MIMETYPE,
     jupyter_download_role,
+    CellOutputsToNodes,
 )
 from .execute import JupyterKernel, ExecuteJupyterCells
 from .thebelab import ThebeButton, ThebeButtonNode, ThebeOutputNode, ThebeSourceNode
@@ -34,6 +39,11 @@ logger = logging.getLogger(__name__)
 def skip(self, node):
     raise docutils.nodes.SkipNode
 
+# Used for nodes that should be gone by rendering time (OutputMimeBundleNode)
+def halt(self, node):
+    raise ExtensionError((f"Rendering encountered a node type that should "
+                         "have been removed before rendering: {type(node)}"))
+
 # Renders the children of a container
 render_container = (
     lambda self, node: self.visit_container(node),
@@ -45,14 +55,17 @@ def visit_container_html(self, node):
     self.body.append(node.visit_html())
     self.visit_container(node)
 
+
 def depart_container_html(self, node):
     self.depart_container(node)
     self.body.append(node.depart_html())
+
 
 # Used to render an element node as HTML
 def visit_element_html(self, node):
     self.body.append(node.html())
     raise docutils.nodes.SkipNode
+
 
 # Used to render the ThebeSourceNode conditionally for non-HTML builders
 def visit_thebe_source(self, node):
@@ -60,6 +73,7 @@ def visit_thebe_source(self, node):
         raise docutils.nodes.SkipNode
     else:
         self.visit_container(node)
+
 
 render_thebe_source = (
     visit_thebe_source,
@@ -170,15 +184,28 @@ def setup(app):
         man=(skip, None),
     )
 
-    # JupyterCellNode is a container that holds the input and
-    # any output, so we render it as a container.
+    # Register our container nodes, these should behave just like a regular container
+    for node in [JupyterCellNode, CellInputNode, CellOutputNode]:
+        app.add_node(
+            node,
+            override=True,
+            html=(render_container),
+            latex=(render_container),
+            textinfo=(render_container),
+            text=(render_container),
+            man=(render_container),
+        )
+
+    # Register the output bundle node.
+    # No translators should touch this node because we'll replace it in a post-transform
     app.add_node(
-        JupyterCellNode,
-        html=render_container,
-        latex=render_container,
-        textinfo=render_container,
-        text=render_container,
-        man=render_container,
+        CellOutputBundleNode,
+        override=True,
+        html=(halt, None),
+        latex=(halt, None),
+        textinfo=(halt, None),
+        text=(halt, None),
+        man=(halt, None),
     )
 
     # JupyterWidgetViewNode holds widget view JSON,
@@ -242,6 +269,7 @@ def setup(app):
     app.add_role("jupyter-download:notebook", jupyter_download_role)
     app.add_role("jupyter-download:script", jupyter_download_role)
     app.add_transform(ExecuteJupyterCells)
+    app.add_post_transform(CellOutputsToNodes)
 
     # For syntax highlighting
     app.add_lexer("ipythontb", IPythonTracebackLexer())
--- a/jupyter_sphinx/__init__.py
+++ b/jupyter_sphinx/__init__.py
@@ -39,10 +39,16 @@ logger = logging.getLogger(__name__)
 def skip(self, node):
     raise docutils.nodes.SkipNode
 
+
 # Used for nodes that should be gone by rendering time (OutputMimeBundleNode)
 def halt(self, node):
-    raise ExtensionError((f"Rendering encountered a node type that should "
-                         "have been removed before rendering: {type(node)}"))
+    raise ExtensionError(
+        (
+            "Rendering encountered a node type that should "
+            "have been removed before rendering: %s" % type(node)
+        )
+    )
+
 
 # Renders the children of a container
 render_container = (
--- a/jupyter_sphinx/ast.py
+++ b/jupyter_sphinx/ast.py
@@ -363,8 +363,8 @@ def attach_outputs(output_nodes, node, thebe_config, cm_language):
     if not node.attributes["hide_code"]:  # only add css if code is displayed
         node.attributes["classes"] = ["jupyter_container"]
 
-    input_node = _return_first_node_type(node, CellInputNode)
-    outputbundle_node = _return_first_node_type(node, CellOutputBundleNode)
+    input_node = _find_only_child_by_type(node, CellInputNode)
+    outputbundle_node = _find_only_child_by_type(node, CellOutputBundleNode)
     output_node = CellOutputNode(classes=["cell_output"])
     if thebe_config:
         # Move the source from the input node into the thebe_source node
@@ -425,7 +425,7 @@ class CellOutputsToNodes(SphinxTransform):
         thebe_config = self.config.jupyter_sphinx_thebelab_config
 
         for cell_node in self.document.traverse(JupyterCellNode):
-            output_bundle_node = _return_first_node_type(
+            output_bundle_node = _find_only_child_by_type(
                 cell_node, CellOutputBundleNode
             )
             # Create doctree nodes for cell outputs.
@@ -451,7 +451,7 @@ class CellOutputsToNodes(SphinxTransform):
             col.process_doc(self.app, node)
 
 
-def _return_first_node_type(node, node_type):
+def _find_only_child_by_type(node, node_type):
     found_nodes = list(node.traverse(node_type))
     if len(found_nodes) == 0:
         raise ValueError("Found no nodes of type %s in node %s" % (node_type, node))
--- a/jupyter_sphinx/ast.py
+++ b/jupyter_sphinx/ast.py
@@ -303,9 +303,6 @@ def cell_output_to_nodes(outputs, data_priority, write_stderr, dir, thebe_config
                 continue
             data = output["data"][mime_type]
             if mime_type.startswith("image"):
-                ####################################
-                # TODO: Figure out how to handle either inline or absolute image paths
-
                 # Sphinx treats absolute paths as being rooted at the source
                 # directory, so make a relative path, which Sphinx treats
                 # as being relative to the current working directory.
@@ -361,10 +358,11 @@ def cell_output_to_nodes(outputs, data_priority, write_stderr, dir, thebe_config
 
 def attach_outputs(output_nodes, node, thebe_config, cm_language):
     if not node.attributes["hide_code"]:  # only add css if code is displayed
-        node.attributes["classes"] = ["jupyter_container"]
+        classes = node.attributes.get("classes", [])
+        classes += ["jupyter_container"]
 
-    input_node = _find_only_child_by_type(node, CellInputNode)
-    outputbundle_node = _find_only_child_by_type(node, CellOutputBundleNode)
+    (input_node,) = node.traverse(CellInputNode)
+    (outputbundle_node,) = node.traverse(CellOutputBundleNode)
     output_node = CellOutputNode(classes=["cell_output"])
     if thebe_config:
         # Move the source from the input node into the thebe_source node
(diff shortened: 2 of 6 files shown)


jupyter_sphinx/__init__.py
```diff
@@ -170,15 +179,28 @@ def setup(app):
...
     app.add_node(
-        JupyterCellNode,
-        html=render_container,
-        latex=render_container,
-        textinfo=render_container,
-        text=render_container,
-        man=render_container,
+        CellOutputBundleNode,
+        override=True,
+        html=(skip, None),
```
akhmerov: Would it make sense to add a `raise` transform that would raise a sphinx extension error, for catching unexpected cases when the transform didn't run?

choldgraf: I'm not sure what a raise translator is, but my intuition is that this would halt the build if a builder that wasn't supported was used? I feel like that'd be extreme no?

akhmerov: This would be to clearly signal, that the node isn't meant to appear in the finalized AST. Seems to make sense and might provide an early notification when something goes awry.

choldgraf: Ahh good point, OK I get it. Would a `raise` translator be something built-in to sphinx? or something we define?

akhmerov: I think it'd be similar to `skip`, which we define in the same file:

```python
def halt(self, node):
    raise sphinx.errors.ExtensionError('Unexpected node encountered ...')
```

I don't expect that sphinx itself would define this because we should be raising an extension error, I think.

choldgraf: ok, I think the latest push adds this, let me know if that's not what you had in mind.

What are your thoughts about testing?

chrisjsewell:
> Need to figure out how to get the testing working. Currently I believe it assumes that the code is executed at parse time, as opposed to in a post-transform. Anybody have an idea for how to do this? My sphinx+pytest skills are not great

See https://github.com/ExecutableBookProject/MyST-Parser/tree/master/tests/test_sphinx for how I test against *minimal* sphinx projects

choldgraf:
@chrisjsewell do you think that a cleaner way to test this, rather than relying on a bunch of one-off assert statements, would be to set up regression tests against the doctree generated by Sphinx? 

And if so, I guess that's what this is doing https://github.com/ExecutableBookProject/MyST-Parser/blob/e37dd0058cba23827b717a4ad0e902c096b8a94f/tests/test_sphinx/conftest.py#L92 ?

Is the idea that you first generate the "correct" XML, and then after that when the tests are run they just compare against that?

chrisjsewell:
> Is the idea that you first generate the "correct" XML, and then after that when the tests are run they just compare against that?

Yes that's correct [pytest-regressions](https://pytest-regressions.readthedocs.io/en/latest/) generates an output the first time you run `pytest`, then tests against that output in all subsequent runs (it can be regenerated with `pytest force-regen`)

Note that the stored doctree XML is the one saved after transforms, but **before** post-transforms. That's why it would also be necessary to test against the created HTML: https://github.com/ExecutableBookProject/MyST-Parser/blob/e37dd0058cba23827b717a4ad0e902c096b8a94f/tests/test_sphinx/conftest.py#L84 (unless there is a way to recover the doctree after post-transforms)

jupyter_sphinx/css/jupyter-sphinx.css
```diff
@@ -57,7 +57,7 @@ div.jupyter_container {
 div.jupyter_container div.highlight {
   background-color: #f7f7f7; /* for haiku */
 }
-div.jupyter_container {
+div.jupyter_container * {
```
akhmerov, line 60: Is this change accidental? I think it reverts a recent CSS bugfix.

jupyter_sphinx/ast.py
```diff
@@ -358,3 +400,50 @@ def get_widgets(notebook):
...
+        for node in self.document.traverse(image):
+            # If the image node has `candidates` then it's already been processed
+            # as in-line markdown, so skip it
+            if "candidates" in node:
+                continue
+            col = ImageCollector()
+            col.process_doc(self.app, node)
+
+
+def _return_first_node_type(node, node_type):
```
akhmerov, line 438: Perhaps `find_only_child_by_type`?

At the same time I'm wondering if using the unpacking idiom could be better than introducing this function. e.g.
`output_bundle_node, = cell_node.traverse(CellOutputBundleNode)`.

What do you think?

jupyter_sphinx/__init__.py
akhmerov: Thanks, this looks good.

I think raising an error when that node is attached is largely an internal consistency check, more akin to an assertion. Therefore I think it is appropriate that it is not covered by tests.

akhmerov:
About the testing strategy: @choldgraf isn't this a separate issue?

akhmerov:
@choldgraf do you understand why the tests fail?

tests/test_execute.py
```diff
@@ -38,7 +38,7 @@ def doctree(source, config=None, return_warnings=False, entrypoint="jupyter_sphi
         apps.append(app)
         app.build()
 
-        doctree = app.env.get_doctree("contents")
+        doctree = app.env.get_and_resolve_doctree("contents", app.builder)
```
akhmerov, line 41: @choldgraf following a sphinx code dive I figured that this should also apply the post-transforms.

@chrisjsewell does this make sense?

akhmerov:
The tests are still failing, but now the that's just because the AST changed and tests require an update.

jupyter_sphinx/ast.py
```diff
@@ -358,3 +400,50 @@ def get_widgets(notebook):
...
+                sphinx_abs_dir(self.env),
+                thebe_config,
+            )
+            # Remove the outputbundlenode and we'll attach the outputs next
+            attach_outputs(output_nodes, cell_node, thebe_config, cell_node.cm_language)
+
+        # Image collect extra nodes from cell outputs that we need to process
+        for node in self.document.traverse(image):
+            # If the image node has `candidates` then it's already been processed
+            # as in-line markdown, so skip it
```
akhmerov, line 431: Not sure what you mean here with "inline markdown". Can you please explain?

jupyter_sphinx/ast.py
```diff
@@ -358,3 +400,50 @@ def get_widgets(notebook):
...
+            if "candidates" in node:
+                continue
+            col = ImageCollector()
```
akhmerov, line 434: I'm a bit concerned that we might be not respecting the app settings here; could we reuse the image collector from the app somehow? I imagine different builders will have different image collectors.

choldgraf: hmm that is a good question. It hasn't been an issue in the Jupyter Book project so far (and we've built both PDF as well as HTML outputs), though I agree it'd be better if we could reuse the image collector that the app uses

akhmerov: I'm also OK leaving this as is and waiting for someone to report a problem.

Please add an extra comment why this is necessary though.

jupyter_sphinx/css/jupyter-sphinx.css
choldgraf: probably yeah, nice catch

tests/test_execute.py
choldgraf: do you mean calling this function also applies post-transforms?

jupyter_sphinx/ast.py
choldgraf: TBH I am not sure, I will try and spend a bit of time early this week to look through this again

tests/test_execute.py
akhmerov: Yep. I verified that after this modification of the fixture, the produces doctrees are correct.

jupyter_sphinx/ast.py
akhmerov: From the processing around I thought you meant that the image what in the document source already. It's just the "inline markdown" that got me confused.

tests/test_execute.py
choldgraf: huh, interesting...so, does that mean we need to change something or that it's OK as-is?

akhmerov: The failures that I see now are due to the AST change (there's now `CellInputNode` containing a literal block and not just a literal block), and not due to the post-transform not being applied.

jupyter_sphinx/__init__.py
```diff
@@ -34,6 +39,11 @@
 def skip(self, node):
     raise docutils.nodes.SkipNode
 
+# Used for nodes that should be gone by rendering time (OutputMimeBundleNode)
+def halt(self, node):
+    raise ExtensionError((f"Rendering encountered a node type that should "
```
akhmerov, line 44: We list python 3.5 as the minimal requirement, so introducing f-strings also requires bumping the python version in setup and in travis config.

I personally don't mind bumping python to 3.6, but @choldgraf you know better the situation with its support across the board.

CC also @SylvainCorlay in case he has objectsions.

tests/test_execute.py
akhmerov: ...so unless I'm overlooking something, this should be good.

jupyter_sphinx/ast.py
choldgraf: yeah I think I probably just used "markdown" because it was coming from myst-markdown, but the overall principle applies, I'll remove the use of "markdown"

jupyter_sphinx/__init__.py
choldgraf: I'm fine removing the f-string because I agree that deprecating 3.5 is a different issue than this PR addresses. My general strategy is to support the latest-3 versions of Python, unless there is a *really* good reason not to

choldgraf:
Just wanna give a heartbeat here - I didn't forget about this PR, just making a push on a "beta" release for Jupyter Book so most of my dev time is directed towards that

akhmerov:
Sure thing, looking forward to the beta release :rocket:

choldgraf:
ok just rebased and pushed to get the new testing infra working on this PR

choldgraf:
ok I think that I've addressed all your comments, but now I remember why this one had slowed down - almost all of the tests are now failing because the structure of the doctrees are hard-coded in there...I can take a stab at converting them over but I hadn't had the time to do it yet

akhmerov:
Before also the post transform didn't run, now it's only the ast that changed.

tests/test_execute/test_basic.xml
```diff
@@ -0,0 +1,5 @@
+<JupyterCellNode classes="jupyter_cell" code_below="False" emphasize_lines="" hide_code="False" hide_output="False" linenos="False" linenostart="True" raises="True" stderr="False">
```
choldgraf, line 1: here's an example of the file that pytest-regressions uses to decide whether it should fail or not. in subsequent runs, if running the test results in a *different* structure than this XML file, then the test will fail

there is one file created per call to `file_regression.check()`

tests/test_execute.py
```diff
@@ -58,36 +58,26 @@ def doctree(
...
     """
     tree = doctree(source, entrypoint="jupyter_sphinx.execute")
     (cell,) = tree.traverse(JupyterCellNode)
-    assert cell.attributes["code_below"] is False
-    assert cell.attributes["hide_code"] is False
-    assert cell.attributes["hide_output"] is False
-    assert cell.attributes["linenos"] is False
-    assert cell.children[0].rawsource.strip() == "2 + 2"
-    assert cell.children[1].rawsource.strip() == "4"
+    file_regression.check(cell.pformat(), extension=".xml")
```
choldgraf, line 80: here is how we use `file_regression`. we pass it some data, and it checks the content of that data against a file that is written to disk in the repository. If they don't match, then the test will fail

choldgraf:
@akhmerov take a look at the latest push, it tests out `pytest-regressions` on two tests. I am curious if you think this would be a good pattern to use for our tests here...that way we can test the full XML structure of the `CellNode` objects, rather than picking out individual pieces from them to test manually.

See a couple github comments I made for what is happening under the hood

chrisjsewell:
Just a reminder; in this PR or another we also want to move this out of myst-nb: https://github.com/executablebooks/MyST-NB/blob/d72ebe707d9ca307b4479c07e3eeed0deba6b996/myst_nb/transform.py#L68

choldgraf:
@chrisjsewell ah right...is this just an inline function for the same thing, basically? And if so could we re-use the same base function for both?

akhmerov:
Re: pytest-regressions,

Yes, appears to be more practical than the current approach of describing the same data in Python.

choldgraf:
note - after thinking about it, I think we should start off by just making sure tests pass here. Then in another PR, we can do a pytest-regressions switch to make sure that the tests already pass as expected

tests/test_execute/test_basic.xml
seanpue: This looks very cool. I wonder though if it would be easier if the XML were included in the test file itself (read as text stream) rather than in separate files. What do you think?

chrisjsewell:
> .is this just an inline function for the same thing, basically?

yep, that we use for the inline glue functionality

> And if so could we re-use the same base function for both?

Quite possiblly, whatever works best 😁

choldgraf:
ok, the tests are now fixed on my end, let's see if they pass. If so, then I'm +1 on a merge and quick release. Then we can finish up the inline function (new issue here: https://github.com/jupyter/jupyter-sphinx/issues/123) and the pytest-regressions move (#122) in another PR to avoid delaying this one further.

jupyter_sphinx/ast.py
```diff
@@ -127,20 +129,24 @@ def run(self):
...
+        cell_node = JupyterCellNode(
+            hide_code=("hide-code" in self.options),
+            hide_output=("hide-output" in self.options),
+            code_below=("code-below" in self.options),
+            linenos=("linenos" in self.options),
+            linenostart=(self.options.get("lineno-start")),
+            emphasize_lines=hl_lines,
+            raises=self.options.get("raises"),
+            stderr=("stderr" in self.options),
+            classes=["jupyter_cell"],
```
akhmerov, line 142: This class is introduced by the PR, right? That seems like it would warrant documenting [here](https://jupyter-sphinx.readthedocs.io/en/latest/#styling-options).

jupyter_sphinx/ast.py
```diff
@@ -127,20 +129,24 @@ def run(self):
...
+        )
+
+        # Add the input section of the cell, we'll add output at execution time
+        cell_input = CellInputNode(classes=["cell_input"])
```
akhmerov, line 146: Same remark about classes.

jupyter_sphinx/ast.py
```diff
@@ -270,6 +303,9 @@ def cell_output_to_nodes(cell, data_priority, write_stderr, dir, thebe_config):
                 continue
             data = output["data"][mime_type]
             if mime_type.startswith("image"):
+                ####################################
+                # TODO: Figure out how to handle either inline or absolute image paths
```
akhmerov, line 307: Which paths are you referring to over here? I thought that we're in full control of where we're writing the image here.

jupyter_sphinx/ast.py
```diff
@@ -373,3 +414,52 @@ def get_widgets(notebook):
...
+
+    default_priority = 700
+
+    def apply(self):
+        thebe_config = self.config.jupyter_sphinx_thebelab_config
+
+        for cell_node in self.document.traverse(JupyterCellNode):
+            output_bundle_node = _find_only_child_by_type(
+                cell_node, CellOutputBundleNode
+            )
```
akhmerov, line 430: I propose to replace the function call with
```suggestion
            output_bundle_node, = cell_node.traverse(CellOutputBundleNode)
```
and similarly in the other two uses of the function `_find_only_child_by_type`.

The unpacking idiom is uncommon, and therefore has a chance to be confusing. But at the same time this implementation saves checking what the function actually does, and to an expert reader not familiar with the code base will be more understandable.

akhmerov:
@choldgraf aside from the remaining comments, everything looks good.

jupyter_sphinx/ast.py
choldgraf: I think this is a carry over from a previous bug? It works fine now in tests and such

choldgraf:
OK, I think I addressed all your comments, WDYT @akhmerov ?

choldgraf:
(also, once you merge I am happy to try bumping the version to see if our new auto-deploy works)

akhmerov:
Thanks, all looks good to merge.

0.2.4 sounds good: all the changes should be backwards compatible, and there's useful new functionality.

Issue #110, opened 2020-04-10
jupyter_sphinx_continue_linenos ignores jupyter_sphinx_linenos setting


SeanDS:
Setting the following in `conf.py` results in line numbers appearing even though `jupyter_sphinx_linenos` is explicitly `False`.

```python
jupyter_sphinx_linenos = False
jupyter_sphinx_continue_linenos = True
```

Seems like it's caused by [this line](https://github.com/jupyter/jupyter-sphinx/blob/3597f9a92f4cf0e766aefc92088f671c319c7186/jupyter_sphinx/execute.py#L184) in `execute.py`:

```python
if linenos_config or continue_linenos or node["linenos"]:
    source["linenos"] = True
```

Maybe it should be the following?

```python
if linenos_config or node["linenos"]:
    source["linenos"] = True
```

akhmerov:
Thanks for reporting. Can you please also elaborate: what is the use case for this combination of options?

SeanDS:
Nothing in particular, just that I wanted to disable line numbering in my docs temporarily, and found it was not actually disabled by setting `jupyter_sphinx_linenos = False`. That seems like weird behaviour to me.

seanpue:
Hi @SeanDS ``linenos`` is a directive flag option in the Sphinx code block, so it would be best to keep it the same.  ``jupyter_sphinx_continue_linenos`` holds the config variable.

I had set the config option ``continue-linenos`` as turning on the line numbering.  Your logic makes more sense.

@akhmerov I am also in favor of adding the ``lineno-start`` directive option, which allows starting at a particular line. That will allow reseting of continuous line numbering by using ``lineno-start: 1``, which is a Sphinx code block directive. I can make these adjustments if it sounds good.

akhmerov:
@seanpue sounds like a good idea to add `lineno-start`.

seanpue:
@akhmerov  I'm implementing this now. One question: Should emphasize-lines match up with the specific linenos? I'm guessing yes.

akhmerov:
Tough call: if the line numbers are continued, then this becomes error prone because the outcome changes if the code above this cell is restructured. Such nonlocal errors are nearly impossible to catch. Therefore I think the current choice is better, even though what you're proposing would be more logical.

seanpue:
Actually @akhmerov never mind. It looks like Sphinx does it relative to the first line as well, so we can hold off on that.

jbweston:
[...]

PR #113, opened 2020-04-20
Fix for #110 and added :lineno-start: directive

seanpue:
Adds :lineno-start: directive and turns off line numbering if continuous line numbering is set.


akhmerov:
All looks good, @seanpue is the syntax change across unrelated tests due to using black or something like that?

seanpue:
Great, @akhmerov Yeah, it's black. I can undo it if you like.

akhmerov:
Black is fine, I'll merge.

PR #114, opened 2020-04-26
Checking if the image path is inside a subdir of dir

AakashGfude:
Related to the discussion here, https://github.com/ExecutableBookProject/MyST-NB/pull/165

cc: @choldgraf @akhmerov

### DIFF
--- a/jupyter_sphinx/ast.py
+++ b/jupyter_sphinx/ast.py
@@ -275,7 +275,7 @@ def cell_output_to_nodes(cell, data_priority, write_stderr, dir, thebe_config):
                 # as being relative to the current working directory.
                 filename = os.path.basename(output.metadata["filenames"][mime_type])
 
-                # checks if file dir path is inside a subpath of dir
+                # checks if file dir path is inside a subdir of dir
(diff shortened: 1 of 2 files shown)


akhmerov:
Thanks! Do you have an idea whether and how this could be tested?

AakashGfude:
@akhmerov sure. I will go through the test suite of this repo and see if I can add something targeting this.

AakashGfude:
@akhmerov @choldgraf what do you think about the test added ?

akhmerov:
The text looks OK to me. It doesn't verify that sphinx with latex would actually work, but neither do the tests right now.

AakashGfude:
@akhmerov I just narrowed down the test to check this particular thing irrelevant to any builder. 
I was thinking of checking the complete latex build, but then I will have to make edits to the `doctree` fixture, to take an array of `sources` and option to create multiple files in different paths. 
If this is something you don't mind having, then we can do it later.

akhmerov:
@AakashGfude, I agree with your choice. I think it's all good to go.

PR #117, opened 2020-04-29
adding github action to publish to pypi

choldgraf:
This is a lightweight github action to publish a repository to PyPI when a new tag is pushed by cutting a release on GitHub. It also adds instructions to follow when cutting a release

It will be hard to test until we actually cut a new release, so maybe we should just try it and see how it goes? What do you think @akhmerov ?

Note: I am also happy to move tests from Travis to GitHub actions, which would standardize the infra a bit more, but only if y'all think it's a good idea

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,27 @@
+name: continuous-integration
+
+on: [push, pull_request]
+
+jobs:
+
+  publish:
+
+    name: Publish to PyPi
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Build package
+        run: |
+          pip install wheel
+          python setup.py sdist bdist_wheel
+      - name: Publish
+        uses: pypa/gh-action-pypi-publish@v1.1.0
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_KEY }}
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,35 @@
+# Release instructions for jupyter-sphinx
+
+Jupyter Sphinx uses a GitHub action to automatically push a new release to
+PyPI when a GitHub release is added.
+
+To cut a new Jupyter Sphinx release, follow these steps:
+
+* Ensure that all tests are passing on master.
+* In [`_version.py`](https://github.com/jupyter/jupyter-sphinx/blob/master/jupyter_sphinx/_version.py),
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,27 @@
...
+          python-version: 3.7
+      - name: Build package
+        run: |
+          pip install wheel
+          python setup.py sdist bdist_wheel
+      - name: Publish
+        uses: pypa/gh-action-pypi-publish@v1.1.0
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_KEY }}
```
choldgraf, line 27: I've added this secret to the repo as well

RELEASE.md
```diff
@@ -0,0 +1,34 @@
...
+  version_info = (0, 2, 3, "final")
+  ```
+* Make a release commit and push to master
+  
+  ```
+  git add jupyter_sphinx/_version.py
+  git commit -m "RLS: 0.2.3"
+  git push upstream master
+  ```
+* [Create a new github release](https://github.com/jupyter/jupyter-sphinx/releases/new).
```
akhmerov, line 22: Is this the same as making an annotated tag of a commit and pushing the tag? If yes: do you think it would be worth to mention?

choldgraf: yep - just pushed an updated explanation

akhmerov:
Any idea what's with the CircleCI error?

choldgraf:
it's just that the config hadn't been merged yet in this PR, so that error should be meaningless

akhmerov:
Does the filename (`tests.yml`) have to be that way? If it may be arbitrary, perhaps rename it to something like `publish`?

choldgraf:
well, I called it `tests.yml` in the hopes that one day we could do testing there instead of with travis :-)  it would be an easy fix, and that way we could ensure tests pass before running the `publish` action

akhmerov:
Got it, thanks.

akhmerov:
Sorry, one final question: right now the skipping of publication will show on any PR. Would it be possible to only apply this action to tag push events and not to PRs?

choldgraf:
Hmm - the way to do that would be to trigger it only on `push` instead of also on `pull_request`, but that's not what we'd want if we were running tests with this action.

If it is helpful, I could run the testing suite in the github action as well, and then compare how it looks vs. travis?

akhmerov:
Can the workflow of publishing be separated from testing (and with different trigger conditions)?

choldgraf:
yes, but then it is difficult to make the "publish" step *depend* on the "test" step passing, which seems like something we'd want, no?

akhmerov:
OK, got it. In that case I think the current solution seems good.

akhmerov:
Also :+1: on moving tests to actions.

maartenbreddels:
You could think of doing a test release on testpypi, or even doing nighly
releases on that :)

(from mobile phone)

PR #119, opened 2020-04-29
adding circleci docs building

choldgraf:
This adds a CircleCI job that builds the jupyter-sphinx documentation so that we can preview it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,40 @@
+version: 2
+jobs:
+  build_docs:
+    docker:
+      - image: circleci/python:3.6-stretch
+    steps:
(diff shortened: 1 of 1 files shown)


choldgraf:
Here's the docs preview: https://8-73843754-gh.circle-artifacts.com/0/html/index.html

akhmerov:
Nice! Is there an easy way to get the URL without going via build logs?

choldgraf:
the artifacts redirector action should do this...I'm not sure why it isn't showing up in this PR, but maybe circle needs to be working on Master before it will work? We can try merging this and I can debug if need be

akhmerov:
Sounds good, merging.

PR #121, opened 2020-04-29
pytests with gha

choldgraf:
Adds a github action to run tests, and makes `publish` depend on it

Currently it doesn't remove the Travis builds (though these are redundant with the GitHub Actions in this PR)...happy to do that if you'd like


akhmerov:
> happy to do that if you'd like

Please go ahead.

akhmerov:
In the logs the actual running of tests is for some reason included in `install dependencies`—that doesn't seem right.

choldgraf:
Yeah - I meant to ask you about that. The reason I did this is because I can't seem to figure out how to get `conda` environments to persist across steps in a job. I haven't used `conda` within my testing infrastructure before so maybe I'm just being a noob?

(as an aside, is there a strong reason to use `conda` in testing for this? it feels a bit heavy-handed given that all the dependencies are on PyPI, no?)

akhmerov:
I'm not well versed with github actions; docs seem to say that one needs to explicitly pass artifacts. There's no strong reason for conda if everything is pip-installable.

One quick workaround would be to just rename step into `test`

choldgraf:
OK, the latest pushes do the following:

* Removes `environment.yml` and replaces it with `requirements.txt`
* Updates the tests to use this instead
* Removes the travis tests

Here's what the docs look like:

https://21-73843754-gh.circle-artifacts.com/0/html/index.html

@akhmerov what do you think?

.github/workflows/tests.yml
```diff
@@ -4,9 +4,33 @@ on: [push, pull_request]
 
 jobs:
 
+  tests:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7]
```
akhmerov, line 12:
```suggestion
        python-version: [3.5, 3.6, 3.7, 3.8]
```

akhmerov:
Thanks, all LGTM

choldgraf:
wohoo - how do you feel about a patch release so we can test if the infrastructure works? I see we already have `0.2.4a1` so maybe it's time for just `0.2.4`? Unless there's something you really wanted in the `.4` patch?

akhmerov:
How about the transform PR? I thought it was nearly done, and iirc it was
overall useful.

choldgraf:
@akhmerov lemme take a quick look

PR #124, opened 2020-05-06
explain how to use hide-code in correctness checks

### LINKED ISSUE #104: Consider a hide-cell option
In a recent thread, the following question came up:

*Is it possible to test documentation made with jupyter-sphinx?*

I realized that an easy way to do this would be to insert `jupyter-execute` cells that ran assert statements and typical testing stuff. However, in that case you wouldn't want the actual code to show up in the final docs. Since there is a `hide-input` and `hide-output` option, what about a `hide-cell` option? Alternatively, if both `hide-input` and `hide-output` are given, this could trigger the behavior that *nothing* makes it into the Sphinx AST (and forgive me if this is already what happens!)

akhmerov:
closes #104 

@choldgraf something like this?

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -147,7 +147,16 @@ produces:
 
     print('this code is invisible')
 
-or vice versa with ``:hide-output:``::
+this option is particularly useful if you want to embed correctness checks in building your documentation::
+
+  .. jupyter-execute::
+      :hide-code:
+
+      assert everything_works, "There's a bug somewhere"
+
+This way even though the code won't make it into the documentation, the build will fail if running the code fails.
+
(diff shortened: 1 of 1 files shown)


doc/source/index.rst
```diff
@@ -147,7 +147,14 @@ produces:
...
     print('this code is invisible')
 
-or vice versa with ``:hide-output:``::
+this option is particularly useful if you want to embed correctness checks in building your documentation::
+
+  .. jupyter-execute::
+      :hide-code:
+
+      assert everything_works, "There's a bug somewhere"
+
```
choldgraf, line 156: I'd add something like "even though the code won't make it into your documentation, your documentation build will error if running the code fails."

PR #125, opened 2020-05-09
force utf8 encoding in write_notebook_output

phaustin:
I'm getting encoding errors in myst_nb without this patch

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -256,7 +256,8 @@ def write_notebook_output(notebook, output_dir, notebook_name):
         resources,
         os.path.join(output_dir, notebook_name + ".ipynb"),
     )
-    # Write a script too.
+    # Write a script too.  Note that utf-8 is the de facto
+    # standard encoding for notebooks. 
     ext = notebook.metadata.language_info.file_extension
     contents = "\n\n".join(cell.source for cell in notebook.cells)
     with open(os.path.join(output_dir, notebook_name + ext), "w",
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -259,7 +259,8 @@ def write_notebook_output(notebook, output_dir, notebook_name):
     # Write a script too.
     ext = notebook.metadata.language_info.file_extension
     contents = "\n\n".join(cell.source for cell in notebook.cells)
-    with open(os.path.join(output_dir, notebook_name + ext), "w") as f:
+    with open(os.path.join(output_dir, notebook_name + ext), "w",
+              encoding = "utf8") as f:
         f.write(contents)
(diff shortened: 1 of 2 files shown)


seanpue:
Hi @phaustin Could you post the encoding error you're getting? When does it happen?

seanpue:
Hi @phausrtin never mind found it!  I wonder if this should try and read an environmental variable with utf8 as the default, just in case someone needs a differ encoding.

akhmerov:
@seanpue good point, I would like to understand the error better: it seems that we're writing a unicode file anyway and not a binary one. I thought one doesn't need to specify encoding in that case.

I'm also wondering if `pathlib.Path.write_text` handles this better.

phaustin:
pinging @choldgraf, @akhmerov and @seanpue -- what do you think about a release with this patch?  This would align the jupyter-sphinx behavior with the rest of jupyter (notebook, nbconvert, jupyterlab).  While sphinx does allow configurable encoding, it's pretty compicated (grepping on "encoding" in the sphinx codebase gives more than 200 hits).   To get the PYTHONUTF8=1 alternative working I think we would have to call sphinx or jupyter-book as a subprocess with that environment variable set, which seems like a fairly egregious hack.

choldgraf:
I think I'm +1 on this idea, if only just to be explicit and to guarantee the same behavior across platforms. As we've learned the "implicit" encoding can be different for windows vs. *nix, so I think we should just set it

phaustin:
perhaps with a warning like jupytext:  https://github.com/mwouts/jupytext/blob/master/jupytext/header.py#L164-L167

phaustin:
I guess https://www.cp1252.com/ ?
On the jupyter side, perhaps there's a writeup somewhere, but as far as the code goes:

https://github.com/jupyter/nbformat/blob/master/nbformat/sign.py#L262-L281

phaustin:
@choldgraf -- once we repeat this for jupyter-cache that should be the end of windows problems for myst_nb

akhmerov:
> https://github.com/jupyter/nbformat/blob/master/nbformat/sign.py#L262-L281

This seems to be about signing in-memory notebooks. This part seems to be more affirmative:  https://github.com/jupyter/nbformat/blob/master/nbformat/__init__.py#L140

That, however, is just the default, and from what I can tell, nbformat would work also with otherwise encoded notebooks. I also see that by default nbformat also forces utf-8 encoding on the saved files.

I've opened an issue in nbformat proposing to clarify this, but for now I am reasonably convinced that the proposed fix is appropriate (although I still don't understand all its implications).

akhmerov:
@phaustin please add a comment that would explain that utf-8 is a de facto standard encoding for notebooks, and I think this is good to go.

PR #127, opened 2020-05-15
Handle missing notebook metadata `file_extension`

chrisjsewell:
The `file_extension` metadata is added to the notebook by the kernel, during execution.
This means that, in cases where the notebook has failed to execute, this line can raise an exception and kill the whole build, for example executablebooks/MyST-NB#175.
If this happens, we would like this line to fail gracefully; either by just not dumping the script file or (as proposed here) defaulting to a specific extension.

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -257,15 +257,14 @@ def write_notebook_output(notebook, output_dir, notebook_name):
         os.path.join(output_dir, notebook_name + ".ipynb"),
     )
     # Write a script too.
-    if ntbk.metadata.get("language_info", {}).get("file_extension", None) is None:
+    ext = notebook.metadata.get("language_info", {}).get("file_extension", None)
+    if ext is None:
         ext = ".txt"
         js.logger.warning(
             "Notebook code has no file extension metadata, " "defaulting to `.txt`",
             # TODO correct location
             # location=document.settings.env.docname,
         )
-    else:
-        ext = notebook.metadata.language_info.file_extension
     contents = "\n\n".join(cell.source for cell in notebook.cells)
     with open(os.path.join(output_dir, notebook_name + ext), "w") as f:
         f.write(contents)
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -206,7 +206,7 @@ class ExecuteJupyterCells(SphinxTransform):
             # Write certain cell outputs (e.g. images) to separate files, and
             # modify the metadata of the associated cells in 'notebook' to
             # include the path to the output file.
-            write_notebook_output(notebook, output_dir, file_name)
+            write_notebook_output(notebook, output_dir, file_name, self.env.docname)
 
             try:
                 cm_language = notebook.metadata.language_info.codemirror_mode.name
@@ -239,7 +239,7 @@ def execute_cells(kernel_name, cells, execute_kwargs):
     return notebook
 
 
-def write_notebook_output(notebook, output_dir, notebook_name):
+def write_notebook_output(notebook, output_dir, notebook_name, location=None):
     """Extract output from notebook cells and write to files in output_dir.
 
     This also modifies 'notebook' in-place, adding metadata to each cell that
@@ -263,8 +263,7 @@ def write_notebook_output(notebook, output_dir, notebook_name):
         ext = ".txt"
         js.logger.warning(
             "Notebook code has no file extension metadata, " "defaulting to `.txt`",
-            # TODO correct location
-            # location=document.settings.env.docname,
+            location=location,
         )
     contents = "\n\n".join(cell.source for cell in notebook.cells)
     with open(os.path.join(output_dir, notebook_name + ext), "w",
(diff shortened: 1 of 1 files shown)


jupyter_sphinx/execute.py
```diff
@@ -257,7 +257,15 @@ def write_notebook_output(notebook, output_dir, notebook_name):
...
-    ext = notebook.metadata.language_info.file_extension
+    if ntbk.metadata.get("language_info", {}).get("file_extension", None) is None:
+        ext = ".txt"
+        js.logger.warning(
+            "Notebook code has no file extension metadata, " "defaulting to `.txt`",
+            # TODO correct location
+            # location=document.settings.env.docname,
+        )
+    else:
+        ext = notebook.metadata.language_info.file_extension
```
akhmerov, line 268:
```suggestion
    ext = notebook.metadata.language_info.file_extension
    if ext is None:
        ext = ".txt"
        js.logger.warning(
            "Notebook code has no file extension metadata, " "defaulting to `.txt`",
            # TODO correct location
            # location=document.settings.env.docname,
        )
```

chrisjsewell: `notebook.metadata.language_info.file_extension` doesn't return None if missing though; it raises an `AttributeError` (See executablebooks/MyST-NB#152)

akhmerov: Sorry, fixed. My multitasking isn't superb.

choldgraf:
oops, sorry, accidental close! I meant to ask where we were on this one...are we just waiting on inserting the location for the warning? If so, I'd be +1 on just merging this since it fixes some downstream bugs (I'm not sure how to test this, but that would be good to)

jupyter_sphinx/execute.py
```diff
@@ -257,7 +257,14 @@ def write_notebook_output(notebook, output_dir, notebook_name):
...
     )
     # Write a script too.
-    ext = notebook.metadata.language_info.file_extension
+    ext = notebook.metadata.get("language_info", {}).get("file_extension", None)
+    if ext is None:
+        ext = ".txt"
+        js.logger.warning(
+            "Notebook code has no file extension metadata, " "defaulting to `.txt`",
+            # TODO correct location
+            # location=document.settings.env.docname,
```
chrisjsewell, line 266: @choldgraf if you can check what the correct code is here, then I'm fine with it

choldgraf: well the docname is sort-of already available to you, but jupyter-sphinx splits it up into the `output_dir` and the `notebook_name` arguments. `output_dir` will be the path to the output dir + any folders in which the current page is nested. and `notebook_name` is just the `basename` of the current document.

So you could either:

1. Split `output_dir` on the path to the root build folder, and take the folders that come after to pre-pend to `notebook_name` and that would be the correct document
2. Make the `output_dir` argument be a path to the root of the build folder, and the `notebook_name` to be the path relative to the root, and then you've already got the `docname` in the function
3. Explicitly pass `docname` as a new argument to `write_notebook_outputs`

akhmerov:
Having read a bit more about the upstream bug, it seems this addresses a use case where:

- `jupyter-sphinx` code tries to save notebooks
- code isn't executed

Under which conditions does this happen? Or are you using some specific components of jupyter-sphinx as a library?

chrisjsewell:
> Or are you using some specific components of jupyter-sphinx as a library?

Well yes in MyST-NB we are using `jupyter-sphinx` as a library, but I guess it may still happen using  `jupyter-sphinx` directly. Basically, the only metadata you actually need to have in a notebook, such that it may run, is the kernel name to load. The kernel itself will populate the rest of the metadata specified in the jsonschema. In fact more specifically, `language_info.file_extension` is not a `required` property of the notebook:  https://github.com/jupyter/nbformat/blob/a523c7189310698dd49fed94e7d84d2946ee4064/nbformat/v4/nbformat.v4.4.schema.json#L31.
If, for whatever reason, the kernel fails to run/populate the notebook, then  `language_info.file_extension` may not be present (unless it has specifically been added).
In this case jupyter_spinx should fail gracefully (just emitting a warning) rather than raising an obscure exception that halts the entire build.

chrisjsewell:
can this be merged @akhmerov?

akhmerov:
@chrisjsewell I thought there is an unresolved conversation and a TODO. What's the status of that?

jupyter_sphinx/execute.py
chrisjsewell: done in 41e945a

chrisjsewell:
> @chrisjsewell I thought there is an unresolved conversation and a TODO. What's the status of that?

oh yeh done 👍

akhmerov:
Thanks!

One last point: you add notebook validation and patching to the code that saves it, which seems to be outside of that function's concern. Wouldn't validation of the metadata be better used outside of that function and before it is called? Or are you using the function directly in myst-nb?

chrisjsewell:
> Wouldn't validation of the metadata be better used outside of that function and before it is called?

well it is validated, by nbformat when it is read. But as mentioned, `language_info.file_extension` is not specifically required by the validating jsonschema in nbformat. So it is then this functions responsibility to account for if it is missing.

akhmerov:
OK, sounds reasonable.

Issue #123, opened 2020-05-06
Add an inline function for cell_outputs_to_nodes


choldgraf:
In MyST-NB we had to write another version of `cell_outputs_to_nodes` to handle inline use-cases: https://github.com/executablebooks/MyST-NB/blob/d72ebe707d9ca307b4479c07e3eeed0deba6b996/myst_nb/transform.py#L68

we should port that over here and re-use as much of the code as possible from the function in jupyter-sphinx

The only differences that I could find were in the kinds of outputs nodes:

* `literal_block` becomes `literal`: https://github.com/executablebooks/MyST-NB/blob/d72ebe707d9ca307b4479c07e3eeed0deba6b996/myst_nb/transform.py#L106
* `container` objects aren't used in the output nodes: https://github.com/executablebooks/MyST-NB/blob/d72ebe707d9ca307b4479c07e3eeed0deba6b996/myst_nb/transform.py#L105

that might be it - but basically it's just making sure the output nodes are single-line-friendly if there's an `inline` flag.

@chrisjsewell to your knowledge what is the difference between the inline one and the one in here? Is it the kinds of nodes that are output? And if so, maybe we just need an "inline=True" flag for the current function?

akhmerov:
Linking a related issue: #40

PR #130, opened 2020-06-23
adding inline functionality to nodes function

choldgraf:
This is a first pass at upstreaming the inline nodes functionality from myst-nb. Definitely might be some bugs in there, I just tried comparing the code between this and https://github.com/executablebooks/MyST-NB/blob/master/myst_nb/transform.py#L71 to make sure that it accomplished the same.

Also updated a test so that we try out this functionality.

cc @chrisjsewell does this seem correct to you?

closes https://github.com/jupyter/jupyter-sphinx/issues/123

### DIFF
--- a/jupyter_sphinx/ast.py
+++ b/jupyter_sphinx/ast.py
@@ -227,7 +227,8 @@ class JupyterWidgetStateNode(docutils.nodes.Element):
         )
 
 
-def cell_output_to_nodes(outputs, data_priority, write_stderr, dir, thebe_config):
+def cell_output_to_nodes(outputs, data_priority, write_stderr, dir,
+                         thebe_config, inline=False):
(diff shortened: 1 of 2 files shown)


choldgraf:
Hmmm, I'm not sure why the tests aren't passing they seem to be working on my machine...any idea @akhmerov ? https://github.com/jupyter/jupyter-sphinx/pull/130/checks?check_run_id=800403700#step:5:32

akhmerov:
At a glance LGTM, I think we can go ahead and merge after a confirmation from @chrisjsewell

choldgraf:
sweet - yeah the real test will be if we can delete the MyST-NB `inline nodes` function and replace it with this, and have all the tests pass there :-)

akhmerov (approved):

Issue #140, opened 2020-07-21
Pre-run cells for intensive computation


willsimmons1465:
I am currently adding jupyter-sphinx to the docs for a project where some methods are especially expensive to run (e.g. connecting to a slow cloud service for running a job) or not feasible to run on our CI runners (e.g. not wanting to use private API keys and credits for the cloud service just for CI and docs building), and would prefer being able to just run it once locally and use the output.

I could not see such this feature available from looking around the docs or looking through the scripts in this repo, but if this is already possible that would be great. This is already available in nbsphinx, but I would prefer to see this in jupyter-sphinx for the added benefit of working inline from the .rst files.

Feature requested: insert the output from locally running the contents of the cell into the .rst file and have it display as if the cell had been run, using the same formatting as usual.

Possible implementation: a `:dont-run:` directive to skip the execution of the cell and just pass it straight to the cell formatting

Temporary hack idea: the first with the code using `:hide-output:` and `:raises:` to accept that it will throw an error when there is no API key found, and the second with just `print(...)` and `:hide-code:` to simulate the output. This would still mean the expensive calls will be made when I am just trying to build the docs locally, which kinda sucks.

akhmerov:
Thanks for the suggestion!

I'd like to understand the use case and the possible implementation better. From your description, it seems that you want a part of the documentation to be computable, but not when building documentation.

If you were using current tools, you'd need to store this data somewhere (likely in the repository), and then use either of the two options:

- Write a jupyter-sphinx cell that only loads and renders the output
- Actually include the output in the .rst using standard tools

What are the disadvantages of these approaches, and how would an alternative look like?

willsimmons1465:
> I'd like to understand the use case and the possible implementation better. From your description, it seems that you want a part of the documentation to be computable, but not when building documentation.

Yes. The particular operations in question are accessing a public access quantum computer to run jobs on it, where supply of credits for the quantum computer are limited and expensive, and access is typically shared with other users (I don't particularly want our regular docs builds to clog up the job queues on these machines and slow down actual important research use). So any solution that would allow us to demonstrate how to submit these jobs but doesn't actually run them during docs builds is good.

> If you were using current tools, you'd need to store this data somewhere (likely in the repository), and then use either of the two options:
> 
> * Write a jupyter-sphinx cell that only loads and renders the output
> * Actually include the output in the .rst using standard tools
> 
> What are the disadvantages of these approaches, and how would an alternative look like?

[...]

PR #141, opened 2020-07-23
Allow adding raw inputs and outputs

willsimmons1465:
Proposed solution for issue #140 

~~Adds an extra option `dont-run` to cells which means they are not submitted for execution, just simply rendered as a cell.~~
~~Current drawback of this solution is that the cell is not included in the .ipynb or .py output files.~~
Adds two new directives, `jupyter-input` and `jupyter-output` which build a `CellInputNode` or `CellOutputNode` directly. These are added to the .ipynb and .py output files when built.
Comments and suggestions welcome.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -305,6 +305,43 @@ produces:
 
     print("hello, world!", file=sys.stderr)
 
+Advanced usage: manually forming Jupyter cells
+----------------------------------------------
+
+When showing code samples that are computationally expensive, access restricted resources, or have non-deterministic output, it can be preferable to not have them run every time you build. You can simply embed input code without executing it using the ``jupyter-input`` directive expected output with ``jupyter-output``::
+
+  .. jupyter-input::
+      :linenos:
+
+      import time
+
+      def slow_print(str):
+          time.sleep(4000)    # Simulate an expensive process
+          print(str)
+    
+      slow_print("hello, world!")
+
+  .. jupyter-output::
+
+      hello, world!
+
+produces:
+
+.. jupyter-input::
+    :linenos:
+
+    import time
+
+    def slow_print(str):
+        time.sleep(4000)    # Simulate an expensive process
+        print(str)
+
+    slow_print("hello, world!")
+
+.. jupyter-output::
+
+    hello, world!
+
 Controlling the execution environment
 -------------------------------------
 The execution environment can be controlled by using the ``jupyter-kernel`` directive. This directive takes
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -317,8 +317,8 @@ produces:
 
     print("hello, world!", file=sys.stderr)
 
-Advanced usage: manually forming Jupyter cells
-----------------------------------------------
+Manually forming Jupyter cells
+------------------------------
 
 When showing code samples that are computationally expensive, access restricted resources, or have non-deterministic output, it can be preferable to not have them run every time you build. You can simply embed input code without executing it using the ``jupyter-input`` directive expected output with ``jupyter-output``::
 
--- a/jupyter_sphinx/ast.py
+++ b/jupyter_sphinx/ast.py
@@ -27,6 +27,51 @@ def csv_option(s):
     return [p.strip() for p in s.split(",")] if s else []
 
 
+def load_content(cell, location, logger):
+    if cell.arguments:
+        # As per 'sphinx.directives.code.LiteralInclude'
+        env = cell.state.document.settings.env
+        rel_filename, filename = env.relfn2path(cell.arguments[0])
+        env.note_dependency(rel_filename)
+        if cell.content:
+            logger.warning(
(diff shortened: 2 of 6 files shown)


akhmerov:
Hm, unfortunately the fix is not as simple, check out the resulting docs: https://86-73843754-gh.circle-artifacts.com/0/html/index.html

The problem is that in order for the styles to be correctly applied, everything should be enclosed in a jupyter cell container.

So in addition to that, we need to make a `JupyterCell` directive, which would create a marked `JupyterCellNode` that has its execution skipped. Optimally, we'd also need to attach the input part of that node in the correct place of the notebook, but only *after* execution, so that the code isn't run. This, however is optional.

Additionally I suggest to rename the docs section to something like "Advanced usage: manually forming Jupyter cells"

willsimmons1465:
@akhmerov Found it easier to just start from master again. Got the formatting mostly there, but can't get the provided inputs and outputs to appear in the same `JupyterCellNode`. If that's what you are talking about, then I think I might have to design it with nested directives, e.g. having `jupyter-input` and `jupyter-output` within a `jupyter-cell` directive

jupyter_sphinx/execute.py
```diff
@@ -149,7 +149,7 @@ def apply(self):
 
             notebook = execute_cells(
                 kernel_name,
-                [nbformat.v4.new_code_cell(node.astext()) for node in nodes],
+                [nbformat.v4.new_code_cell(node.astext() if node["execute"] else "") for node in nodes],
```
akhmerov, line 152: This will have a downside of the cells that aren't executed not available for download in a generated notebook or a script. Should we improve on that by:

- Remembering at which locations these cells would be inserted.
- Reinserting them into the notebook after the execution has finished (and also handling the outputs at that point)

akhmerov: If such a thing existed, we could mark the cells with "please don't execute tag", but it doesn't exist.

jupyter_sphinx/ast.py
```diff
@@ -152,6 +163,136 @@ def run(self):
         cell_node += cell_input
         return [cell_node]
 
+class InputCell(Directive):
```
akhmerov, line 166: I think `CellInput` (and likewise `CellOutput`) are more appropriate names because both refer to an input and an output *of a cell*.

jupyter_sphinx/execute.py
willsimmons1465: This is already done - see the other change made to execute.py. The insertion of an empty cell here means that zipping the nodes and cells of the notebook still keeps them aligned correctly and we can hence replace the empty cells with the given input/output code as needed

akhmerov: Ah, great, sorry that I didn't catch that first! Please leave a comment here about the purpose of the empty code cells.

I was also for a second concerned that this would cause a hit in performance due to pinging kernel with empty messages, but it seems nbclient already skips executing empty code cells.

akhmerov:
Sorry, took me a while to come back to this.

I have looked at the implementation and all looks good except for two related aspects:

- I think it should be an error to use a `jupyter-output` directive without an immediately preceding `jupyter-input`, let's catch this case and raise an extension error.
- Probably related: the outputs and inputs seem to be contained within different jupyter cell containers, which results in a different rendering (a gap between the input and the output). Let's instead detect when they are neighbors in the doctree and combine them into a single one. Let me know if you would benefit from advice on the implementation.

image

doc/source/index.rst
```diff
@@ -305,6 +305,43 @@ produces:
 
     print("hello, world!", file=sys.stderr)
 
+Advanced usage: manually forming Jupyter cells
+----------------------------------------------
```
akhmerov, line 309:
```suggestion
Manually forming Jupyter cells
------------------------------
```

willsimmons1465:
> * Probably related: the outputs and inputs seem to be contained within different jupyter cell containers, which results in a different rendering (a gap between the input and the output). Let's instead detect when they are neighbors in the doctree and combine them into a single one. Let me know if you would benefit from advice on the implementation.

I have added an extra pre-processing transform to combine the `JupyterCellNode`s from input and outputs when adjacent. The formatting looks much more consistent now which is nice

akhmerov:
Yes, this is great!

Looking at the tests I see that we assume that output without input is an error, and don't assume that input without output is an error. This is appropriate, however let's make it official and add a test that *only* `jupyter-input` does work and produces a cell without output.

After that I think all good to go!

willsimmons1465:
Because a `CellOutputBundleNode` is always added to each jupyter cell (see [this line](https://github.com/jupyter/jupyter-sphinx/blob/b37e5516f0e3c6b1d3583e9052ef3d386477b813/jupyter_sphinx/execute.py#L244)), we will still have an output cell when only `jupyter-input` is given, but we can check that it is empty and hidden

akhmerov (approved):

akhmerov:
Awesome, thanks for your work @willsimmons1465!

PR #147, opened 2020-08-08
Fix Windows CI

### LINKED ISSUE #54: maint: use pathlib
Since we're relying on Python >=3.5, we may as well use the glory of pathlib.

### LINKED ISSUE #143: Potential issue with sphinx_abs_dir on Windows
Hey @akhmerov why are you breaking our tests 😜 

Copied from https://github.com/executablebooks/jupyter-book/pull/723#issuecomment-670207614:

Looks like the error is raised from `jupyter_sphinx` when the sphinx app builds. It tries to get the `relpath` between the src and out directories, where the out directory is a temp folder on the drive that `pytest` is installed on.

You can't generate relative paths between Windows drives without creating symbolic links, so either:
1. the temp test outputs would need to be written on the same drive as the book src
2. or the **absolute** out path would need to be passed to sphinx (though it looks like sphinx expects an absolute path)

Trying 2: `test_toc_startwithlist` and `test_toc_startwithheader` pass when I edit [`sphinx_abs_dir()` from `jupyter_sphinx`](https://github.com/jupyter/jupyter-sphinx/blob/46f2ea3b432b8eb263e5da67038d45043903647d/jupyter_sphinx/utils.py#L77) to return the absolute out dir instead, using:
```
return output_directory(env)
```

so it might be feasible to do this when a Windows path is detected?

foster999:
- [x] Non-posix paths are causing CI to fail on Windows. Converted to posix

- [x] `sphinx_abs_dir` tries to get the relative path between two Windows drives when testing. Changed to use absolute path on Windows

- [x] tests check that  `sphinx_abs_dir` output is relative path using `assert_node`. Updated `test_download_role` to check if Windows absolute path is equivalent to the relative reftarget.

Fixes #143, fixes #54
Should fix some issues in Windows tests at executablebooks/jupyter-book#723

### DIFF
--- a/jupyter_sphinx/ast.py
+++ b/jupyter_sphinx/ast.py
@@ -229,7 +229,7 @@ class JupyterWidgetStateNode(docutils.nodes.Element):
         )
 
 
-def cell_output_to_nodes(outputs, data_priority, write_stderr, dir, thebe_config):
+def cell_output_to_nodes(outputs, data_priority, write_stderr, out_dir, thebe_config):
     """Convert a jupyter cell with outputs and filenames to doctree nodes.
 
     Parameters
@@ -239,7 +239,7 @@ def cell_output_to_nodes(outputs, data_priority, write_stderr, dir, thebe_config
         Which media types to prioritize.
     write_stderr : bool
         If True include stderr in cell output
-    dir : string
+    out_dir : string
         Sphinx "absolute path" to the output folder, so it is a relative path
         to the source folder prefixed with ``/``.
     thebe_config: dict
@@ -305,19 +305,17 @@ def cell_output_to_nodes(outputs, data_priority, write_stderr, dir, thebe_config
                 continue
             data = output["data"][mime_type]
             if mime_type.startswith("image"):
+                file_path = Path(output.metadata["filenames"][mime_type])
+                out_dir = Path(out_dir)
                 # Sphinx treats absolute paths as being rooted at the source
                 # directory, so make a relative path, which Sphinx treats
                 # as being relative to the current working directory.
-                filename = os.path.basename(output.metadata["filenames"][mime_type])
+                filename = file_path.name
 
-                # checks if file dir path is inside a subdir of dir
-                filedir = os.path.dirname(output.metadata["filenames"][mime_type])
-                subpaths = filedir.split(dir)
-                if subpaths and len(subpaths) > 1:
-                    subpath = subpaths[1]
-                    dir += subpath
+                if out_dir in file_path.parents:
+                    out_dir = file_path.parent
 
-                uri = Path(os.path.join(dir, filename)).as_posix()
+                uri = (out_dir / filename).as_posix()
                 to_add.append(docutils.nodes.image(uri=uri))
             elif mime_type == "text/html":
                 to_add.append(
--- a/jupyter_sphinx/ast.py
+++ b/jupyter_sphinx/ast.py
@@ -102,8 +102,7 @@ class JupyterCell(Directive):
                     location=location,
                 )
             try:
-                with open(filename) as f:
-                    content = [line.rstrip() for line in f.readlines()]
+                content = [line.rstrip() for line in Path(filename).read_text()]
             except (IOError, OSError):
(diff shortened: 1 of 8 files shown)


jupyter_sphinx/ast.py
```diff
@@ -316,7 +317,7 @@ def cell_output_to_nodes(outputs, data_priority, write_stderr, dir, thebe_config
                     subpath = subpaths[1]
                     dir += subpath
 
-                uri = os.path.join(dir, filename)
+                uri = Path(os.path.join(dir, filename)).as_posix()
```
akhmerov, line 320: perhaps

```suggestion
                uri = (Path(dir) / filename).as_posix()
```

foster999: Will do 👍

jupyter_sphinx/execute.py
```diff
@@ -265,8 +267,10 @@ def write_notebook_output(notebook, output_dir, notebook_name, location=None):
             location=location,
         )
     contents = "\n\n".join(cell.source for cell in notebook.cells)
-    with open(os.path.join(output_dir, notebook_name + ext), "w",
-              encoding = "utf8") as f:
+
+    notebook_file = notebook_name + ext
+    output_dir = Path(output_dir)
+    with (output_dir / notebook_file).open("w", encoding = "utf8") as f:
         f.write(contents)
```
akhmerov, line 274:
```suggestion
    (output_dir / notebook_file).write_text(contents, encoding = "utf8")
```

foster999: Thanks, hadn't spotted that method

akhmerov:
I see the failures in Python 3.5. It's at its EoL in September, so I think we should drop it.

foster999:
> I see the failures in Python 3.5. It's at its EoL in September, so I think we should drop it.

I think they could be genuine fails on Linux paths - I'd only tested locally on Windows py3.8. Will look into it

foster999:
Fixed a couple of other bugs I'd created, but this passes on Ubuntu py3.6. @akhmerov are we able to trigger the other CI checks to double check this, before we drop 3.5?

akhmerov:
It's annoying since the jobs get canceled as soon as one fails. Yet, 3.7 and 3.8 work, and I think so should 3.6.

akhmerov:
So I'd say let's go and drop 3.5.

jupyter_sphinx/utils.py
```diff
@@ -74,12 +75,14 @@ def sphinx_abs_dir(env, *paths):
...
-        os.path.abspath(
-            os.path.join(output_directory(env), os.path.dirname(env.docname), *paths)
-        ),
-        os.path.abspath(env.app.srcdir),
-    )
+    out_path = (output_directory(env) /  Path(env.docname).parent / Path(*paths)).resolve()
+      
+    if os.name == "nt":
+        # Can't get relative path between drives on Windows
+        return out_path.as_posix()
```
akhmerov, line 82: Does this actually produce a download link that works? I can't check on windows, but it seems that the filename has a different structure and I wonder if sphinx resolves this correctly.

tests/test_execute.py
```diff
@@ -34,16 +35,17 @@ def doctree():
...
-        src_dir = tempfile.mkdtemp()
-        source_trees.append(src_dir)
-        with open(os.path.join(src_dir, "conf.py"), "w", encoding = "utf8") as f:
-            f.write("extensions = ['%s']" % entrypoint)
-            if config is not None:
-                f.write("\n" + config)
-        with open(os.path.join(src_dir, "contents.rst"), "w", encoding = "utf8") as f:
-            f.write(source)
+        src_dir = Path(tempfile.mkdtemp())
+        source_trees.append(src_dir.as_posix())
```
akhmerov, line 39: I think this is sufficient:

```suggestion
        source_trees.append(src_dir)
```

jupyter_sphinx/utils.py
foster999: Pleasantly surprised that it works locally and when [hosted](https://foster999.github.io/test-jupyter-sphinx-win/)

PR #155, opened 2020-08-27
Fix handling of math in outputs

### LINKED ISSUE #134: Incorrect handling of mathjax
While moving execution to a post-transform (#107) brought a lot of advantages, it also causes problems, and this is one of those.

As far as I understand, sphinx determines whether to include mathjax in a page before applying post-transforms. This may result in mathjax missing from the build if there are no regular math directives, see e.g. https://59-73843754-gh.circle-artifacts.com/0/html/index.html

akhmerov:
Domain resolution (and in particular math domain resolution) happens before post_transforms. Therefore it seems like the only option to correctly include mathjax math is to move including outputs from a post transform to a transform, like I do in this PR.

Closes #134

I'm unsure about the implications and appropriateness of this though. The tests still pass and the docs look good.

@chrisjsewell can you please give some advice/explanation here? Is there a downside I'm overlooking?

PS this is the last PR in 0.3 milestone aside of cleanup ;)

### DIFF
--- a/jupyter_sphinx/__init__.py
+++ b/jupyter_sphinx/__init__.py
@@ -272,7 +272,7 @@ def setup(app):
     app.add_role("jupyter-download:nb", JupyterDownloadRole())
     app.add_role("jupyter-download:script", JupyterDownloadRole())
     app.add_transform(ExecuteJupyterCells)
-    app.add_post_transform(CellOutputsToNodes)
+    app.add_transform(CellOutputsToNodes)
 
     # For syntax highlighting
     app.add_lexer("ipythontb", IPythonTracebackLexer)
--- a/tests/test_execute.py
+++ b/tests/test_execute.py
@@ -46,7 +46,7 @@ def doctree():
         if config is not None:
             conf_contents += "\n" + config
         (src_dir / "conf.py").write_text(conf_contents, encoding = "utf8")
-        (src_dir / "contents.rst").write_text(source, encoding = "utf8")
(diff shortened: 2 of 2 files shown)


chrisjsewell:
Well again, this transform doesn't affect myst-nb, so I have no strong opinion lol.
But the general point of not doing it in a transform, as opposed to a post-transform, is that now the doctree you store will not be builder independent. So there could be issues trying to build e.g. HTML and LaTeX without having to rebuild

Somewhat related, I re-process the std domain here: https://github.com/executablebooks/MyST-NB/blob/a8fa6c029605ec16cc28f520464fae13b8164681/myst_nb/render_outputs.py#L448-L458

akhmerov:
> Well again, this transform doesn't affect myst-nb, so I have no strong opinion lol.

Yes, I'm not asking because of compatibility. Purely selfish reasons.

> Somewhat related, I re-process the std domain here

I considered this option, but is it reliable to just re-process math?

chrisjsewell:
> I considered this option, but is it reliable to just re-process math?

Well you'll have to look at the math domains `process_doc` and see what it does 😉 
Unfortunately its a bit of a pain either way

akhmerov:
Alright, I poked around sphinx source and realized I really don't want to mess with things like reference resolution.

Switching this back to a post-transform if there's a demand for multiple output priorities only takes re-adding 5 symbols, but in the meantime everything will work better.

PR #159, opened 2020-08-28
Fix script outputs

chrisjsewell:
v0.3.2 please lol

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -271,7 +271,7 @@ def write_notebook_output(notebook, output_dir, notebook_name, location=None):
             "Notebook code has no file extension metadata, " "defaulting to `.txt`",
             location=location,
         )
-    contents = "\n\n".join(cell.source for cell in notebook.cells)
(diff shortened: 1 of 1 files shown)


akhmerov:
Good idea. I am wondering: wouldn't it be better to send this via nbconvert script export though?

chrisjsewell:
What extra would that give you?

akhmerov:
AFAIU non-code cells would be left as comments.

chrisjsewell:
How does it no for every programming language though what the comment syntax is?

akhmerov:
Looking at the [source](https://github.com/jupyter/nbconvert/blob/master/nbconvert/exporters/script.py#L33-L36) via an entry point. I presume these would typically be provided by the kernel package, e.g. nbconvert docs specifically mention Julia.

chrisjsewell:
Well if it works, then yeh go for it, just make sure that it falls back on outputting the code only (with the correct file extension)

akhmerov:
Closing in favor of #160

PR #160, opened 2020-08-30
use nbconvert to write scripts

akhmerov:
Compared to the current implementation

- Adds a correct shebang for known langauges
- If a comment format is known, converts markdown to comments
- Falls back gracefully to only include code cells if it doesn't know what to do

@chrisjsewell :point_up: I think this supersedes #159 

I think we could lean more on the nbconvert API, but this seems like a direct incremental improvement over what we have now.

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -1,6 +1,7 @@
 """Execution and managing kernels."""
 
 import os
+import warnings
 from pathlib import Path
 from logging import Logger
 
@@ -18,6 +19,22 @@ if nbconvert.version_info < (6,):
      from nbconvert.preprocessors.execute import executenb
 else:
      from nbclient.client import execute as executenb
+
+import traitlets
+# Workaround of https://github.com/ipython/traitlets/issues/606
+if traitlets.version_info < (5, 1):
+    class LoggerAdapterWrapper(Logger):
+        """Wrap a logger adapter, while pretending to be a logger."""
+        def __init__(self, wrapped):
+            self._wrapped = wrapped
+
+        def __getattribute__(self, attr):
+            if attr == "_wrapped":
+                return object.__getattribute__(self, attr)
+            return self._wrapped.__getattribute__(attr)
+else:
+    def LoggerAdapterWrapper(logger_adapter):
+        return logger_adapter
 
 import nbformat
 
@@ -39,20 +56,6 @@ from .ast import (
     attach_outputs,
     get_widgets,
 )
-
-
-class LoggerAdapterWrapper(Logger):
-    """Wrap a logger adapter, while pretending to be a logger.
-
-    Workaround of https://github.com/ipython/traitlets/issues/606
-    """
-    def __init__(self, wrapped):
-        self._wrapped = wrapped
-
-    def __getattribute__(self, attr):
-        if attr == "_wrapped":
-            return object.__getattribute__(self, attr)
-        return self._wrapped.__getattribute__(attr)
 
 
 class JupyterKernel(Directive):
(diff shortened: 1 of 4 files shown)


jupyter_sphinx/execute.py
```diff
@@ -264,18 +264,12 @@ def write_notebook_output(notebook, output_dir, notebook_name, location=None):
     )
     # Write a script too.  Note that utf-8 is the de facto
     # standard encoding for notebooks. 
-    ext = notebook.metadata.get("language_info", {}).get("file_extension", None)
```
chrisjsewell, line 267: What happens now if this key doesn't exist and/or the kernel does not exist or provide the correct information? Does nbconvert degrade gracefully?

Or should these lines be wrapped in a try/except+logger.warning, to ensure this "non-essential" output does not kill the whole sphinx build if it errors?

chrisjsewell:
Thanks @akhmerov, see the code comment

chrisjsewell:
and looks like issues with the logger 😬

akhmerov:
> and looks like issues with the logger :grimacing:

https://github.com/ipython/traitlets/issues/606 :tired_face:

I'll think what to do about those.

chrisjsewell:
FYI, you may want to try using tox (see https://github.com/executablebooks/MyST-NB/blob/master/tox.ini), which I've found recently makes it super easy to run these tests locally, without the need to set up a development environment

jupyter_sphinx/execute.py
akhmerov: > Does nbconvert degrade gracefully?

Yes. Both if `language_info` is missing and if `language_info = {'name': 'unknown'}`. Annoyingly nbconvert ignores the extension even if it is stored in the metadata. Not sure if that occurs in the wild, and I'm inclined to make it an issue in nbconvert.

akhmerov:
> FYI, you may want to try using tox (see https://github.com/executablebooks/MyST-NB/blob/master/tox.ini), which I've found recently makes it super easy to run these tests locally, without the need to set up a development environment

Sounds good, #161

jupyter_sphinx/execute.py
akhmerov: I was wrong, nbconvert properly treats all metadata.

chrisjsewell: Perhaps also test it against a non-standard kernel? Like in https://github.com/executablebooks/MyST-NB I now build a docs page using the https://github.com/evhub/coconut kernel:

```bash
pip install coconut
pip install ipython<=7.11.0  # required by coconut 1.4.3
coconut --jupyter # register kernel
```

akhmerov:
Still remaining: tests, including coconut kernel, although I gotta avoid making this a dependency.

jupyter_sphinx/execute.py
```diff
@@ -262,20 +262,16 @@ def write_notebook_output(notebook, output_dir, notebook_name, location=None):
...
-        js.logger.warning(
-            "Notebook code has no file extension metadata, " "defaulting to `.txt`",
-            location=location,
-        )
-    contents = "\n\n".join(cell.source for cell in notebook.cells)
 
-    notebook_file = notebook_name + ext
+    # Write a script too. Due to https://github.com/ipython/traitlets/issues/606
+    # we are using sphinx logger directly withou its adapter.    
+    exporter = nbconvert.exporters.ScriptExporter(log=js.logger.logger)
```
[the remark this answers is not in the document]
akhmerov: Very important point. We for sure lose formatting and log context, but indeed I'd need to confirm that -W works correctly.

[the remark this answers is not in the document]
akhmerov: ...or I could have nbconvert and traitlets bow to my power :laughing: 

```
class LoggerWrapper(Logger):
    def __init__(self, wrapped):
        self._wrapped = wrapped

    def __getattr__(self, attr):
        attr = self._wrapped.__getattribute__(attr)

logger = LoggerWrapper(js.logger)
```

jupyter_sphinx/execute.py
akhmerov: I decided to go with bash kernel.

Issue #157, opened 2020-08-27
Fix flaky windows tests


akhmerov:
There's a warning about IOPub timeout and the failures are nondeterministic.

choldgraf:
perhaps the hackiest-but-fastest thing to do is to add a couple of catches to re-run this test if it fails with that error automatically. That way the test would only fail if it fails 3 times in a row or something

akhmerov:
NB This seems to be the same issue: https://github.com/mwouts/jupytext/issues/489

chrisjsewell:
theres a pytest plugin for that: https://github.com/pytest-dev/pytest-rerunfailures

chrisjsewell:
but the more complete fix would be to add an `IOPub timeout` option to parse to nbclient

chrisjsewell:
https://github.com/jupyter/nbclient/blob/09dbc75ac83e3df322a2a9fb614d4c82f40b6fe9/nbclient/client.py#L176

akhmerov:
Didn't seem to help on jupytext side: https://github.com/mwouts/jupytext/issues/489#ref-commit-7d73894

There's a draft PR against nbconvert, that claims to fix this https://github.com/jupyter/nbconvert/pull/1183

PR #166, opened 2020-09-21
🚀 RLS: v0.3.2

choldgraf:
this gets in the bugfix for the `download` role. If @akhmerov is 👍 on this then I'll make a release so this gets pushed to pypi

### DIFF
--- a/jupyter_sphinx/_version.py
+++ b/jupyter_sphinx/_version.py
@@ -1,4 +1,4 @@
-version_info = (0, 3, 1, "final")
+version_info = (0, 3, 2, "final")
 
 _specifier_ = {"alpha": "a", "beta": "b", "candidate": "rc", "final": ""}
 


akhmerov (approved):
Sounds good. I was thinking about other possible fixes, but there's no reason to hold this off.

choldgraf:
yeah benefits of making the release process as easy as possible is we can make more releases in general :-)

choldgraf:
OK if tests pass I will merge and cut a new release

choldgraf:
@akhmerov any idea why the tests would suddenly start failing on windows?

akhmerov:
#147 some weird kernel messaging timeouts, that trigger once in a while.

akhmerov:
Eh, wrong reference, here's the correct issue: #157

choldgraf:
ok the re-boot of the test is happy, so I'm gonna merge!

PR #169, opened 2020-09-27
Allow mime bundles display different content based on the builder

### LINKED ISSUE #156: implement builder-dependent output priority
Right now the output priority is the same for all formats, which puts non-HTML formats at a disadvantage.

Implementing this would require reverting #155 and #158 where we switched output processing from a post-transform to a transform, but that in turn breaks reference processing with e.g. math in outputs, see discussion in #155.

akhmerov:
Closes #156 

So far just a proof of principle, and mainly an RFC.

The main workhorse of the implementation is the `MimeBundleNode`, which pretends being different types of output, based on the the build context.

The advantages of this implementation are:

- I won't need to run transforms or post-transforms (the refactoring still pending)
- This plays nicely with the domain registration (reference collection, searching for math content, indexing images, etc). AFAIK there were a handful of bugs related to that.

Possible disadvantage: all parsing at the build stage is unconditional. So if someone adds a numbered equation that is also represented as a html widget, then the equation number reference would point nowhere in the html docs. I tend to consider this a minor limitation since it's not even clear what the output means semantically (e.g. how should the equation number work in html?)

Result in action:

`make -C doc clean html`

image

`make -C doc latexpdf` (+skip a bunch of unrelated errors from unicode in math)

image

Pinging @choldgraf and @chrisjsewell, what do you think?

### DIFF
--- a/jupyter_sphinx/ast.py
+++ b/jupyter_sphinx/ast.py
@@ -2,6 +2,7 @@
 
 import os
 import json
+import contextlib
 from pathlib import Path
 
 import docutils
@@ -316,12 +317,35 @@ class CellOutputNode(docutils.nodes.container):
         super().__init__("", **attributes)
 
 
-class CellOutputBundleNode(docutils.nodes.container):
-    """Represent a MimeBundle in the Sphinx AST, to be transformed later."""
+class MimeBundleNode(docutils.nodes.container):
+    """A node with multiple representations rendering as the highest priority one."""
 
-    def __init__(self, outputs, rawsource="", *children, **attributes):
-        self.outputs = outputs
-        super().__init__("", **attributes)
+    def __init__(self, rawsource="", *children, **attributes):
+        super().__init__("", *children, mimetypes=attributes["mimetypes"])
+
+    def render_as(self, visitor):
+        """Determine which node to show based on the visitor"""
+        try:
+            # Or should we go to config via the node?
+            priority = visitor.builder.env.app.config[
+                'render_priority_' + visitor.builder.format
+            ]
+        except (AttributeError, KeyError):
+            # Not sure what do to, act as a container and show everything just in case.
+            return super()
+        for mimetype in priority:
+            try:
+                return self.children[self.attributes['mimetypes'].index(mimetype)]
+            except ValueError:
+                pass
+        # Same
+        return super()
+
+    def walk(self, visitor):
+        return self.render_as(visitor).walk(visitor)
+
+    def walkabout(self, visitor):
+        return self.render_as(visitor).walkabout(visitor)
 
 
 class JupyterKernelNode(docutils.nodes.Element):
@@ -372,15 +396,13 @@ class JupyterWidgetStateNode(docutils.nodes.Element):
         )
 
 
-def cell_output_to_nodes(outputs, data_priority, write_stderr, out_dir,
+def cell_output_to_nodes(outputs, write_stderr, out_dir,
                          thebe_config, inline=False):
     """Convert a jupyter cell with outputs and filenames to doctree nodes.
 
     Parameters
     ----------
     outputs : a list of outputs from a Jupyter cell
-    data_priority : list of mime types
-        Which media types to prioritize.
     write_stderr : bool
         If True include stderr in cell output
     out_dir : string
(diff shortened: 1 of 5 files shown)


jupyter_sphinx/ast.py
```diff
@@ -183,6 +185,35 @@ def __init__(self, outputs, rawsource="", *children, **attributes):
...
+                return self.children[self.attributes['mimetypes'].index(mimetype)]
+            except ValueError:
+                pass
+        return super()
+
+    def walk(self, visitor):
+        return self.render_as(visitor).walk(visitor)
+
+    def walkabout(self, visitor):
+        return self.render_as(visitor).walkabout(visitor)
```
akhmerov, line 214: The key part of the implementation.

akhmerov:
I'm also interested if a similar implementation would work for myst-nb.

chrisjsewell:
> I'm also interested if a similar implementation would work for myst-nb.

If you can get https://myst-nb.readthedocs.io/en/latest/use/formatting_outputs.html#images to work without reference duplication warnings.

Note, if you want to go this route, you can most likely delete a lot of this code and just use `sphinx.addnodes.only()`

akhmerov:
> If you can get https://myst-nb.readthedocs.io/en/latest/use/formatting_outputs.html#images to work without reference duplication warnings.

Can you explain a bit more? There's only one image node ever created, why would there be a reference duplication? Do you have a related closed issue or any other place that provides more details?

> Note, if you want to go this route, you can most likely delete a lot of this code and just use sphinx.addnodes.only()

Not sure how: this node combines global configuration and its contents in order to determine how it should be rendered. `only` only seems to have access to global tags. Can you explain how that node would help?

At a glance the `MimeBundleNode` allows to avoid the rendering-related post-transforms, and removes the need to copypaste a lot of sphinx internal code (image processing, math domain indexing, etc).

chrisjsewell:
> Can you explain how that node would help?

Ah fair, its a slightly similar approach, but whereas you are rendering all mimetypes initially, then selecting the specific one later, it renders all build types initially, then selects a specific one.
This is how nbsphinx works.

> Can you explain a bit more?

So a "common" example would be if you set up matplotlib to output both an SVG (for HTML ouputs) and a png/PDF (for TeX outputs).

If you also set the rendering to add a caption/reference, I believe in your example, this will be equivalent to writing something like:

```restructuredtext
.. only: html

   .. figure:: path/to/image.svg
      :name: myfig

      caption

.. only: latex

   .. figure:: path/to/image.pdf
      :name: myfig

      caption
```

this would cause a duplicate reference error

but there might be a way to handle this, by trying to utilise sphinx's asterisks '*' feature; grouping all the image mime-types together and only storing a single representation for all of them, which would be equivalent to:

```restructuredtext
.. figure:: path/to/image.*
   :name: myfig

   caption
```

akhmerov:
Hmm, the figure metadata applies to the whole cell. Wouldn't it be then appropriate to create a single figure node, that would contain a `MimeBundleNode` inside? AFAIK the figure node is a container for an image, so that seems to check out.

There are bizarre corner cases, but I don't know how to meaningfully support an output that wants to be a figure in one format and a numbered equation in another.

chrisjsewell:
> Hmm, the figure metadata applies to the whole cell. Wouldn't it be then appropriate to create a single figure node, that would contain a MimeBundleNode inside?

No I'm dubious that would work

akhmerov:
It's hard to read through sphinx internals, but in the rendering process the typical patter is that nodes don't query their children. Figure node is [not an exclusion](https://github.com/sphinx-doc/sphinx/blob/22820d841e6e794961d58f1b22a1f8362538d83d/sphinx/writers/latex.py#L1294-L1328). Therefore until the rendering actually happens it won't check that its first child is a `MimeBundleNode`, and not an image. When the rendering does happen, `MimeBundleNode` will behave exactly like an image.

akhmerov:
Note also some differences: the new `MimeBundleNode` isn't the same as the `CellOutputBundleNode`: it only represents a single node, unlike cell outputs that may have as many as one wants. I think this makes sense because the only responsibility of the node is to:

- Translate from jupyter mime formats to sphinx nodes
- Display the correct one based on the builder priority.

chrisjsewell:
Oh yeh in general I think it might work. But I'd want to make sure there was rigorous regression testing before merging

akhmerov:
Following the [confirmation](https://groups.google.com/g/sphinx-users/c/3lvql2-GRcI/m/nnkRzWOIAgAJ) from the sphinx devs that the overall approach is :ok_hand:, I'm proceeding to implement this.

To do so far:

- [ ] ensure that both the outputs and display data are treated as mime bundles (right now outputs aren't yet)
- [ ] review the codebase to check for code that is now redundant
- [ ] fix the tests (`jupyter-input` and `jupyter-output` are now failing).
- [ ] add new tests

PR #181, opened 2021-04-23
Fix syntax highlighting logic.

philj56:
Previously, the language attribute was being set on the wrong node,
causing sphinx to use the default Python 3 highlighting for all
jupyter-execute blocks. In addition, the IPython3 lexer was being given
the wrong name. This commit fixes those issues, causing ipython %magic
commands to be highlighted correctly, and allowing non-Python-3 kernels
to use their own highlighting.


akhmerov:
Thanks!

PR #200, opened 2022-04-01
docs: fix typos

kianmeng:


akhmerov:
Thanks! I'm merging because the test failures are unrelated (those are due to jinja2 upgrade).

PR #204, opened 2022-06-25
Do not overwrite files in _static if user already has custom versions of those files

basnijholt:
The CSS files are always copied in `build_finished`, this prevents one from overwriting any of the CSS files.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -413,6 +413,10 @@ Here is a sample ``custom.css`` file overriding the ``stderr`` background color:
   }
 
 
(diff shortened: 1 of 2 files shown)


akhmerov:
I reviewed the code, and this looks good, thanks! Perhaps also add an entry to the changelog, and an explanation to the styling section of the docs.

basnijholt:
I have added a line to the changelog about the bug fix but where would you like me to mention `pre-commit`?

There is no other section about Git or dev tools.

akhmerov:
I would say changelog is user-focused, so we don't need to put pre-commit there. We could add a readme section "Contributing" in a separate PR.

akhmerov:
Excellent, thanks!

PR #206, opened 2022-06-25
Add pre-commit with various linters and static code checkers

basnijholt:
This adds pre-commit with black, isort, pyupgrade, and a few other nice to haves with a new Github Action that runs pre-commit.

The only changes I made manually are in https://github.com/jupyter/jupyter-sphinx/pull/206/commits/7f35bea0a2123d7e7315a4411992751141751d83 (which also include a few automatically made changes by isort).


jupyter_sphinx/ast.py
```diff
@@ -45,8 +41,8 @@ def load_content(cell, location, logger):
         try:
             with Path(filename).open() as f:
                 content = [line.rstrip() for line in f.readlines()]
-        except (IOError, OSError):
-            raise IOError("File {} not found or reading it failed".format(filename))
+        except OSError:
+            raise OSError(f"File {filename} not found or reading it failed")
```
akhmerov, line 45: Why the change to the exception types?

basnijholt: This change is made by pyupgrade and is equivalent because we support Python≥3.6 only.

See https://github.com/asottile/pyupgrade#oserror-aliases

basnijholt: `isinstance(IOError(), OSError) == True`

akhmerov (approved):

PR #207, opened 2022-06-25
Introduce new jupyter-download syntax, fixes #182

### LINKED ISSUE #182: Warning in sphinx 4.0.0
Sphinx released a new version that seems to cause this kind of warnings:

```bash
WARNING: unknown directive or role name: jupyter-download:notebook
WARNING: unknown directive or role name: jupyter-download:script
```

when using something like:

```
:jupyter-download:notebook:`Tutorial 1`
:jupyter-download:script:`Tutorial 1`
```

The docs build seems to be ok with the correct output, might just make some pipelines fail due to the warnings

basnijholt:
Since Sphinx 4.0.0 using `:` inside of a role/directive does not work, see #182.
Therefore, we add `-` as a separator to get, `jupyter-download-notebook`, `jupyter-download-nb`, and `jupyter-download-script`.
We leave the `:` syntax for backward compatibility reasons.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -372,16 +372,16 @@ Downloading the code as a script
 --------------------------------
 
 Jupyter Sphinx includes 2 roles that can be used to download the code embedded in a document:
-``:jupyter-download:script:`` (for a raw script file) and ``:jupyter-download:notebook:`` or ``:jupyter-download:nb:`` (for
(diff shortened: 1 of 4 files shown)


akhmerov:
Overall this looks good. Please add a short update to the 0.4.0 changelog. I think we should also add a deprecation warning to the old roles and promise to eliminate the `:` syntax in 0.5.0.

basnijholt:
I just added a changelog entry and a DeprecationWarning.

Issue #222, opened 2023-03-06
0.4.0: pytest is failing in `tests/test_execute.py::test_save_script` unit because hardcoded python interpreter name


kloczek:
Looks like pytest is failing in `tests/test_execute.py::test_save_script` because hardcoded pytjon interpreter name.
Insterd it should be used `sys.executable`.
pytest as well currently shows some `PendingDeprecationWarning` warnings.
Here is pytest output:
<details>

```console
+ PYTHONPATH=/home/tkloczko/rpmbuild/BUILDROOT/python-jupyter-sphinx-0.4.0-4.fc35.x86_64/usr/lib64/python3.8/site-packages:/home/tkloczko/rpmbuild/BUILDROOT/python-jupyter-sphinx-0.4.0-4.fc35.x86_64/usr/lib/python3.8/site-packages
+ /usr/bin/pytest -ra -m 'not network'
============================= test session starts ==============================
platform linux -- Python 3.8.16, pytest-7.2.2, pluggy-1.0.0
rootdir: /home/tkloczko/rpmbuild/BUILD/jupyter-sphinx-0.4.0
collected 35 items

tests/test_execute.py ...........................Fs......                [100%]

=================================== FAILURES ===================================
_______________________________ test_save_script _______________________________

doctree = <function doctree.<locals>.doctree at 0x7fac32562040>

    def test_save_script(doctree):
        source = """
        .. jupyter-kernel:: python3
          :id: test

        .. jupyter-execute::

          a = 1
          print(a)
        """
        _, app, _ = doctree(source, return_all=True)
        outdir = Path(app.outdir)
        saved_text = (outdir / "../jupyter_execute/test.py").read_text()
>       assert saved_text.startswith("#!/usr/bin/python3")
E       AssertionError: assert False
E        +  where False = <built-in method startswith of str object at 0x7fac33135d30>('#!/usr/bin/python3')
E        +    where <built-in method startswith of str object at 0x7fac33135d30> = '#!/usr/bin/env python\n# coding: utf-8\n\n# In[1]:\n\n\na = 1\nprint(a)\n\n'.startswith

tests/test_execute.py:637: AssertionError
=============================== warnings summary ===============================
../../../../../usr/lib/python3.8/site-packages/jupyter_client/connect.py:27
  /usr/lib/python3.8/site-packages/jupyter_client/connect.py:27: DeprecationWarning: Jupyter is migrating its paths to use standard platformdirs
[...]

PR #226, opened 2023-08-01
Add direct dependency on ipykernel

garrison:
ipywidgets 8.1.0 has been released, which includes the change at https://github.com/jupyter-widgets/ipywidgets/pull/3811. With this change, ipykernel is no longer a dependency of ipywidgets, so jupyer-sphinx must now depend on ipykernel directly rather than assume it will be installed as a transitive dependency of ipywidgets.  Without this change, jupyter-sphinx tests will fail CI. This has additionally led to CI failures in some packages that depend on jupyter-sphinx but not ipykernel.

The minimum version chosen here is the same minimum version that was formerly specified by ipywidgets until https://github.com/jupyter-widgets/ipywidgets/pull/3811.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 sphinx>=2.4.4
+ipykernel>=4.5.1
 ipywidgets>=7.0.0
 IPython
 nbconvert>=5.4
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ setup(
     packages=["jupyter_sphinx"],
     install_requires=[
         "Sphinx>=2",
+        "ipykernel>=4.5.1",
         "ipywidgets>=7.0.0",
         "IPython",
         "nbconvert>=5.5",


bnavigator:
You're editing the wrong file. The relevant dependency definition is in setup.py

https://github.com/jupyter/jupyter-sphinx/blob/1bb7d916d8ade01ce30d74cdf6cf343bec6236ec/setup.py#L31-L37

garrison:
> You're editing the wrong file. The relevant dependency definition is in setup.py

Sorry, I erroneously assumed that this repository would obey DRY and read `requirements.txt` in `setup.py`.  Fixed.

12rambau:
note that it fixes #222

bnavigator:
> note that it fixes #222

The error message is not the same. Your issue was a missing ipykernel for `python3`, the reporter of #222 has a modified script intrepreter line (shebang) because of a non-standard test setup.

mhvk:
It would be nice to get this in - I now added a dependence to `ipykernel` in my package, but ideally `jupyter-sphinx` should be enough.

p.s. Thanks for a really nice extension to sphinx!

PR #237, opened 2023-12-09
Adapt to level config to match other Jupyter repos

blink1073:
- Adopts the `.pre-commit-config.yaml` hooks used by `jupyter_client`, including the Style & static checks from the [Scientific Python Development Guide](https://learn.scientific-python.org/development/guides/style/).
- Switches to `hatch` for build backend and task runner.
- Adds contributing document.
- Removes use of deprecated `.traverse()` in favor of `.findall()`.


tests/test_execute.py
```diff
@@ -539,9 +544,9 @@ def test_latex(doctree):
 
     for start, end in delimiter_pairs:
         tree = doctree(source.format(start, end))
-        (cell,) = tree.traverse(JupyterCellNode)
+        (cell,) = tree.findall(JupyterCellNode)
         (_, celloutput) = cell.children
-        assert next(iter(celloutput.traverse(math_block))).astext() == r"\int"
+        assert next(iter(celloutput.findall(math_block))).astext() == r"\int"
```
akhmerov, line 549: `iter` is unnecessary here, right?

jupyter_sphinx/execute.py
```diff
@@ -123,12 +123,12 @@ def apply(self):
         linenos_config = self.config.jupyter_sphinx_linenos
         continue_linenos = self.config.jupyter_sphinx_continue_linenos
         # Check if we have anything to execute.
-        if not doctree.traverse(JupyterCellNode):
+        if not list(doctree.findall(JupyterCellNode)):
```
akhmerov, line 126: Here and later converting to list looks OK, but I wonder if keeping an iterator is cleaner
```suggestion
        if not next(doctree.findall(JupyterCellNode), False):
```

akhmerov (approved):

tests/test_execute.py
blink1073: Fixed

jupyter_sphinx/execute.py
blink1073: Updated

akhmerov (approved):

.pre-commit-config.yaml
```diff
@@ -1,24 +1,76 @@
...
+  autoupdate_commit_msg: "chore: update pre-commit hooks"
+
 repos:
-  - repo: "https://github.com/pre-commit/pre-commit-hooks"
-    rev: v4.3.0
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.5.0
     hooks:
-      - id: trailing-whitespace
+      - id: check-case-conflict
```
12rambau, line 9: I usually use less pre-commit hooks relying mostly on the combo: black, prettier, ruff, codespell. Is there any advantage of using all these ones ? 
Is it simply mirroring the good practices of Jupyter packages (in this case I'll keep it as it is) ?

pyproject.toml
```diff
@@ -0,0 +1,84 @@
...
+    "nbconvert>=5.5",
+    "nbformat",
+    "Sphinx>=7",
+]
+
+[project.urls]
+"Bug Tracker" = "https://github.com/jupyter/jupyter-sphinx/issues/"
+Documentation = "https://jupyter-sphinx.readthedocs.io"
+Homepage = "https://github.com/jupyter/jupyter-sphinx/"
+"Source Code" = "https://github.com/jupyter/jupyter-sphinx/"
```
12rambau, line 28: why setting up a "source code" page if it's the same as the "HomePage" ? Is it compulsory ?

.pre-commit-config.yaml
blink1073: Yes, we've been copy-pasting pre-commit files for a while now.  :smile:

pyproject.toml
blink1073: These links were directly pulled from `setup.py`, I'll change the HomePage to `jupyter.org`.

12rambau:
I was looking at the repository history, and it seems normal "merge" has been used so far. For this PR only "squash and merge" is available, is it also something you changed @blink1073 ? 

Note that if it's something we can change I am more in favor of a normal merge as it gives more credit to big contributions to the repository than small fixes and as long as the naming convention is correctly managed, it still allow the use of a meaningful changelog (the one generated from Github works just fine)

12rambau:
@blink1073 I have reactivated the regular merge in the settings but I cannot set it back for this specific branch. Is it ok for you if I squash and merge so we can start from here for further development ?

blink1073:
I disabled "Allow merge commits" because they clutter up the history with merges from "main".  Squash merging should be preferred IMO, unless there truly is some reason each commit in a PR has different meaning (which to me means that the PR should have been split).  However, "Rebase merging" can be used in that scenario.

12rambau:
@blink1073 I got an upvote from @akhmerov so I think this matter could be discussed. On my side I still see a huge disadvantage to squash merging: It gives the same recognition to a typo PR and a huge feature PR. As an example, I was dead inside when I realized that these 2 PRs (https://github.com/pydata/pydata-sphinx-theme/pull/540, https://github.com/pydata/pydata-sphinx-theme/pull/962/files) were given the same importance in my contributing section. 

The problem you mention seems to be an egg and chicken issue as the "merge from main" conflict only occurs when you squash the previous PR. If not, git is very good at reconciling everything and Github provides a very convinient "sync" button that does everything for you from any branch. 

Another example, I tried to contribute to a repository where people were less available for reviews (https://github.com/widgetti/ipyvuetify/pulls) but still using the same convention as the one you suggested: small single-feature PR. The consequence is that I have PR sitting there for 6 months preventing me from moving forward as I need them pushed sequentially. As this repository doesn't get as much maintainer love as the other Jupyter repos, I think it's very relevant to allow some motivated contributors to make big PR without squashing them.

blink1073:
I'm afk this week, and will defer to whatever you both think is best for this repo. My main point is that you can still see all the commits with the rebase option, but they are at least grouped.  As long as the no-op "merge" commits aren't in the history I think it is fine.

12rambau:
No worries, I'll pile some small PR in the meantime and wait for your return for the bigger ones. 
I tried to do it myself but I don't have access to branch policies and it seems the blocking parameter is coming from here.

PR #248, opened 2023-12-28
build: use a dedicated publish workflow

12rambau:
- The publish workflow is only triggered from a Github release. 
- It runs the tests from the other workflow "tests.yml" to avoid duplication
-  I refactored a bit the tests.yml file


12rambau:
self-merging as I need these pipelines to run to release v0.5 on pypi

PR #265, opened 2023-12-29
test: use pytest-regression in our test suit

### LINKED ISSUE #122: Use pytest-regressions for tests
pytest-regressions is a useful way to check that some structured data in your tests doesn't change. We could use this in our own tests to make them a bit more complete and easier to parse.

12rambau:
Fix #122


akhmerov:
I have a bit mixed feelings about this approach.

Pros:
- It is less code
- It is easier to add new tests
Cons:
- The tests compare a lot more information (the entire HTML) rather than the doctree. This may make tests fail if e.g. sphinx changes its markup.
- Perhaps more significantly: verifying that the check is correct requires manual audit of the HTML, and is potentially error prone.

I don't think the drawbacks are severe, but I wanted to bring them up for consideration.

12rambau:
I also thought about it but I didn't find efficient ways to parse doctree nodes as a json tree which would have allowed me to continue testing the document object. 

That being said, the final result is the html output and changes from sphinx needs to be captured very early by the tests to avoid breaking custom css and themes so I think it brings lots of advantages as well. It is still possible to use other builders if the html output is not sufficient or if issues are identified in pdf or epub.

12rambau:
@akhmerov do you have any last though on this one before I merge it ? 
I would like to have it merged before I merge other PR to avoid the tests misalignment.

akhmerov:
Sorry for not being clear. I think this is a perfectly fine solution that removes a lot of boilerplate.

PR #268, opened 2023-12-29
chore: update the commit hooks list

12rambau:
As mentioned in #237, I updated the list of pre-commit to something more comfortable. 

The new list is more restricted and cover pretty much the same things as many were overlapping: 
- black for the python file linting (I just aligned the line length to ruff at 100 characters) 
- prettier for anything else (json and yaml schema are tested as well for parsing purposes) 
- nbstripout because we have notebooks
- doc8 (maybe not the best choice but the only one I know for Sphinx checks) 
- ruff, the ligthning fast swiss army knife PEP checker

I also change the lint hatch session so it run the exact same thing as the one in the test suit.

I was forced to add some ignore in ruff for documentation issues.

### DIFF
--- a/jupyter_sphinx/execute.py
+++ b/jupyter_sphinx/execute.py
@@ -186,7 +186,7 @@ class ExecuteJupyterCells(SphinxTransform):
                     if output["output_type"] == "stream" and output["name"] == "stderr"
                 ]
                 if stderr and not node.attributes["stderr"]:
-                    js.logger.warning("Cell printed to stderr:\n{}".format(stderr[0]["text"]))
+                    js.logger.warning(f"Cell printed to stderr:\n{stderr[0]['text']}")
 
             # Insert input/output into placeholders for non-executed cells
             for node, cell in zip(nodes, notebook.cells):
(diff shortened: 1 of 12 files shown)


tests/test_execute.py
```diff
@@ -219,8 +218,7 @@ def test_continue_linenos_conf_option(doctree):
 
     tree = doctree(
         source,
-        config="jupyter_sphinx_linenos = True\n"
-        "jupyter_sphinx_continue_linenos = True",
+        config="jupyter_sphinx_linenos = True\n" "jupyter_sphinx_continue_linenos = True",
```
akhmerov, line 221: Here and below superfluous string break. Other than that all LGTM

jupyter_sphinx/execute.py
```diff
@@ -192,9 +186,7 @@ def apply(self):
                     if output["output_type"] == "stream" and output["name"] == "stderr"
                 ]
                 if stderr and not node.attributes["stderr"]:
-                    js.logger.warning(
-                        "Cell printed to stderr:\n{}".format(stderr[0]["text"])
-                    )
+                    js.logger.warning("Cell printed to stderr:\n{}".format(stderr[0]["text"]))
```
akhmerov, line 189:
```suggestion
                    js.logger.warning(f"Cell printed to stderr:\n{stderr[0]['text']}")
```

akhmerov (approved):

